PR #52, opened 2022-02-02
Utilities related to charset

acabarbaye:
### Description

Added ability to:
- detect a charset from a string
- convert a string from a charset to another

### Test Coverage


SamCHogg (approved):
I don't think I know enough about this problem to know whether or not this is the best solution to the problem. But I can see there is good test coverage, and that although there used to be some go charset libraries these now seem unmaintained.

So I'll approve this, but also provide some additional things I found/thought:
*  I did find this post about one of those libraries where the maintainer talks about the alternative: https://groups.google.com/g/golang-nuts/c/FHmzYmM5r5Y. No clue if that is actually of any help though.
* An alternative approach could not have a pure go implementation and use the C libraries that handle this, which looks to be the approach taken here: https://github.com/uber-go/icu4go. Though this introduces its own challenges.
* Otherwise you could call out to the `unzip` process and let that handle the problem, but that's then not portable and is highly OS specific.

One suggestion I do have is that maybe this should be a separate library, and follow the principal of "do one thing well".

acabarbaye:
golang-utils is meant to be a library which can be included in all kinds of go programs and so the code needs to be portable. This removes the ability to use any of the dependencies out there related to charset as they almost all assume C libraries are installed on the platform and so, portability would be lost.
There are also two aspects of the problem:

    Detect a charset used in a string (for this I use a pure go dependency)
    convert a string to another charset (for this I use go standard library implementation golang.org/x/text/transform)

The problem about charsets is that there is no proper standardisation of the labels used to describe a charset and each tool uses its own labelling. So most of the work was about creating some mapping so that the different bits understand each other

    I did find this post about one of those libraries where the maintainer talks about the alternative: https://groups.google.com/g/golang-nuts/c/FHmzYmM5r5Y.

for the conversion, I use what is suggested there (i.e. go's transform)

    An alternative approach could not have a pure go implementation and use the C libraries that handle this, which looks to be the approach taken here: https://github.com/uber-go/icu4go.

golang-utils needs to ensure portability and cannot make assumptions of what is installed on a machine

    Otherwise you could call out to the unzip process and let that handle the problem, but that's then not portable and is highly OS specific.

every unzip tool work differently and we have specific requirements about unzip (e.g. keep timestamps, etc.). We are not implementing unzip but using what go provides to do so. Nonetheless, we need some extra processing on top to ensure we have a unzip method which behaves exactly the same on any platform.

    One suggestion I do have is that maybe this should be a separate library, and follow the principal of "do one thing well".

Opensourcing libraries is unfortunately not a simple and quick process. We try to keep the scope limited to language utilities though so that the library does not get too big. The fact that it is all in one place makes it easier to maintain.

PR #55, opened 2022-02-02
handle non UTF-8 filenames when unzipping zip files

acabarbaye:
### Description

handle non UTF-8 filenames when unzipping zip files:

If a file in the zip has a filename which is not encoded using UTF-8 (as specified but not enforced by the specification), then we try to detect what is the charset used and if successful, we then convert the filename to UTF-8 using the charset transformation utility.

Zip files used as test fixtures were modified to include files with characters other than ASCII

I also took the opportunity to make `OpenFile` available as a module function.

### Test Coverage


utils/filesystem/files.go
```diff
@@ -185,6 +187,10 @@ func (fs *VFS) GenericOpen(name string) (File, error) {
 	return convertFile(func() (afero.File, error) { return fs.vfs.Open(name) }, func() error { return nil })
 }
 
+func OpenFile(name string, flag int, perm os.FileMode) (File, error) {
+	return globalFileSystem.OpenFile(name, flag, perm)
+}
```
stevep-arm, line 192: Is this function being used?

acabarbaye: not in this library but I think Graham was asking to have access as a filesystem global API

utils/filesystem/files.go
```diff
@@ -1165,6 +1159,34 @@ func (fs *VFS) copyZipFile(dest string, zippedFile *zip.File) (err error) {
...
+	// Currently an error is raised when non-UTF8 characters are present in filepath
+	// See https://go-review.googlesource.com/c/go/+/75592/
+	// Character encodings other than CP-437 and UTF-8
+	//are not officially supported by the ZIP specification, pragmatically
+	//the world has permitted use of them.
+	//
+	//When a non-standard encoding is used, it is the user's responsibility
+	//to ensure that the target system is expecting the encoding used
+	//(e.g., producing a ZIP file you know is used on a Chinese version of Windows).
+	if !utf8.ValidString(destinationPath) {
```
stevep-arm, line 1174: This is a preference thing rather than a requirement but I would write this the other way round.

```
if utf8.ValidString(destinationPath) {
    return desinationPath, nil
}
```

Just to avoid a function where most of it is indented in the if.

utils/filesystem/files.go
acabarbaye: I added a changelog entry to document that change and mentioned it in the PR description

utils/filesystem/files.go
```diff
@@ -1165,6 +1159,34 @@ func (fs *VFS) copyZipFile(dest string, zippedFile *zip.File) (err error) {
 	return
 }
 
+func determineUnzippedFilepath(dest string) (string, error) {
+	destinationPath := dest
```
stevep-arm, line 1163:
```suggestion
func determineUnzippedFilepath(destinationPath string) (string, error) {
```

`dest` doesn't seem to be used so may as well change the name of the argument.

utils/filesystem/files.go
acabarbaye: this is a fair point. I changed the function

stevep-arm (approved):

utils/filesystem/files.go
madchutney: Thanks for adding this, we did want a simple function to call in Core-Tools.

madchutney (approved):
LGTM

PR #61, opened 2022-02-15
Make http clients more flexible

acabarbaye:
### Description
- Added missing mocks
- Made it possible to use default client as IClient
- Improved testing

### Test Coverage


changes/202202151234.feature
```diff
@@ -0,0 +1 @@
+Possibility to create `OPTIONS` requests using an http.IClient
```
joshjennings98, line 1: [According to the way news files should be made](https://github.com/ARM-software/golang-utils/blob/dc840f91f3436c712f2d7355dc2015759db4b649/CONTRIBUTING.md#news-files), they should be YYYYMMDD not YYYYMMDDhhmm and you only need one for the whole merge, I see why you did this and it makes sense but then should the news files notes be updated?

changes/202202151235.feature
```diff
@@ -0,0 +1 @@
+Added missing mocks
```
joshjennings98, line 1: Would this be a `.feature`? Wouldn't adding tests come under:
> Changes to the repository that do not impact functionality
Which would make this a `.misc`

Or maybe a `.bug` if the mocks were meant to be there?

acabarbaye: in most cases, this would be true but in this case the Mocks are exported so that they can be used in projects importing this library. For those projects, new mocks are hence new "features"

joshjennings98: Ah okay

utils/http/interfaces.go
```diff
@@ -35,16 +35,18 @@ type IClient interface {
...
-	Post(url, bodyType string, body interface{}) (*http.Response, error)
+	Post(url, contentType string, body interface{}) (*http.Response, error)
 	// PostForm is a convenience method for doing simple POST operations using
 	// pre-filled url.Values form data.
 	PostForm(url string, data url.Values) (*http.Response, error)
 	// StandardClient returns a standard library *http.Client with a custom Transport layer.
 	StandardClient() *http.Client
 	// Put performs a PUT request.
-	Put(url string, rawBody interface{}) (*http.Response, error)
+	Put(url string, body interface{}) (*http.Response, error)
```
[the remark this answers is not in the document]
acabarbaye: The parameter name does not really matter as it is the signature which is looked at. However, for clarity, I changed the parameter name so that it matches the interface description

joshjennings98 (approved):
Looks good to me :+1:

PR #63, opened 2022-03-01
[Reflection] get the pointer to a value

acabarbaye:
### Description

- Added a way to get the pointer to a value

### Test Coverage

### DIFF
--- a/utils/reflection/reflection_test.go
+++ b/utils/reflection/reflection_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/bxcodec/faker/v3"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


utils/reflection/reflection_test.go
```diff
@@ -487,3 +488,92 @@ func TestIsEmpty(t *testing.T) {
...
+	}
+	tests := []struct {
+		input          interface{}
+		expectedOutput interface{}
+		expectedError  error
+	}{
+		{
+			input:          nil,
+			expectedOutput: nil,
+			expectedError:  commonerrors.ErrUnsupported,
```
joshjennings98, line 518: Are there any other cases that would cause the error or is nil the only one? Maybe something that will cause `!vp.CanInterface` to error and another that causes `!obj.IsValid` to error?

If you pass it a `var v interface{}` it should return an error shouldn't it?

joshjennings98 (commented):
Just a minor note on the test

utils/reflection/reflection_test.go
acabarbaye: I added more test cases for ensuring both conditions are tested.
From what I could see, canInterface only return false for not exposed structure fields https://stackoverflow.com/questions/50279840/when-is-go-reflect-caninterface-false

joshjennings98 (approved):
Looks good to me

PR #113, opened 2022-08-08
Add git utility to prevent cloning git bombs

joshjennings98:
### Description

Add git utility to validate git repos before cloning so we can prevent git bombs.

### Test Coverage


utils/git/limits.go
```diff
@@ -0,0 +1,110 @@
+package git
+
+import (
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+
+	"github.com/ARM-software/golang-utils/utils/config"
+)
+
+const MaxInt64 = int64(^uint64(0) >> 1)
```
acabarbaye, line 9: use math.MaxInt64 instead

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return worktree.Checkout(opts)
+}
+
+func NewCloneObject(cfg RepositoryConfig, path string) (c CloneObject, err error) {
+	err = c.Initialise(path, cfg)
+	return
+}
+
+func CloneWithLimits(ctx context.Context, url, dir string, limits ILimits) (err error) {
```
acabarbaye, line 182: could you add some documentation

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return
+	}
+
+	err = c.ValidateRepository(ctx)
+	if err != nil {
+		return
+	}
+
+	return c.Checkout(&git.CheckoutOptions{Branch: plumbing.ReferenceName("HEAD")})
```
acabarbaye, line 202: could the checkout logic in https://github.com/Arm-Debug/workspace-service-api/blob/063adb24d1ed34226db9a883a5415173ec0faeb4/api/workspace/git_workspace.go#L77 be moved here so that we can checkout the correct branch/tag/hash and apply the general options (e.g. submodules, depth, ...)

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func CloneWithLimits(ctx context.Context, url, dir string, limits ILimits) (err error) {
+	c, err := NewCloneObject(RepositoryConfig{
+		MaxTreeDepth:      limits.GetMaxTreeDepth(),
+		MaxRepositorySize: limits.GetMaxTotalSize(),
+		MaxFileCount:      limits.GetMaxFileCount(),
+		MaxFileSize:       limits.GetMaxFileSize(),
+		MaxEntries:        limits.GetMaxEntries(),
+		CloneOptions: &git.CloneOptions{
```
acabarbaye, line 189: see comment below

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+	MaxEntries        int64
+	Parallelisation   int
+}
+
+type CloneObject struct {
+	cfg  RepositoryConfig
+	repo *git.Repository
+}
+
+func (c *CloneObject) Initialise(path string, cfg RepositoryConfig) (err error) {
```
acabarbaye, line 48: Could this be renamed Clone

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	errs, ctx := errgroup.WithContext(ctx)
+
+	totalSize := atomic.NewInt64(0)
+	totalFileCount := atomic.NewInt64(0)
+	totalEntries := atomic.NewInt64(0)
+
+	var wg sync.WaitGroup
+	for p := 0; p < Parallelisation; p++ {
```
acabarbaye, line 87: could this be split into smaller functions as this is a bit hard to follow

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type CloneObject struct {
+	cfg  RepositoryConfig
+	repo *git.Repository
+}
+
+func (c *CloneObject) Initialise(path string, cfg RepositoryConfig) (err error) {
+	fs := osfs.New(path)
+	storer := memory.NewStorage()
```
acabarbaye, line 50: repository can be really large and we do not want it all in memory

utils/git/git.go
```diff
@@ -0,0 +1,204 @@
...
+		MaxFileCount:      limits.GetMaxFileCount(),
+		MaxFileSize:       limits.GetMaxFileSize(),
+		MaxEntries:        limits.GetMaxEntries(),
+		CloneOptions: &git.CloneOptions{
+			URL: url,
+		},
+	}, dir)
+	if err != nil {
+		return
+	}
```
acabarbaye, line 195: I think I may have missed something but the validation of the repository is done after the cloning so wouldn't the bomb have "exploded" by this point?

joshjennings98: We don't check it out. I tested it manually first. If you do `git clone XXX --no-checkout` then the git bomb doesn't go off. I make sure that the checkout doesn't happen and then run the validation. It is only when you run a git command that the bomb goes off, until then it seems to be equivalent to downloading the repo.

acabarbaye (approved):

PR #121, opened 2022-08-23
⭐ Added `shared cache` using the filesystem

acabarbaye:
### Description

- Added a way to create and manage shared cache using solely the file system

### Test Coverage

### DIFF
--- a/utils/sharedcache/common.go
+++ b/utils/sharedcache/common.go
@@ -1,0 +1,312 @@
+package sharedcache
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/ARM-software/golang-utils/utils/collection"
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+	"github.com/ARM-software/golang-utils/utils/hashing"
+	"github.com/ARM-software/golang-utils/utils/parallelisation"
+	"github.com/ARM-software/golang-utils/utils/reflection"
+)
+
+const defaultCachedPackage = "cache.zip"
+const hashFileDescriptor = ".hash"
+
+// GenerateKey generates a key based on a list of key elements `elems`.
+func GenerateKey(elems ...string) string {
+	hash := "ef46db3751d8e999" // xxhash(""). No point in calculating it for "" everytime
+	for _, elem := range elems {
+		hash = hashing.CalculateHash(hash+elem, hashing.HashXXHash)
+	}
+	return hash
+}
+
+func generateHashFileName(srcFile string) string {
+	return fmt.Sprintf("%v%v", filepath.Base(srcFile), hashFileDescriptor)
+}
+func getHash(ctx context.Context, fs filesystem.FS, src string, forceHashUpdate bool) (hash string, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	hashFile := filepath.Join(filepath.Dir(src), generateHashFileName(src))
+	if fs.Exists(hashFile) && !forceHashUpdate {
+		fileContents, readErr := fs.ReadFile(hashFile)
+		if readErr == nil {
+			hash = string(fileContents)
+			if len(hash) == 16 { // if valid hash
+				// xxhash 64 bit will have 16 characters
+				return
+			}
+		}
+	}
+
+	hash, err = fs.FileHash(hashing.HashXXHash, src)
+	if err != nil {
+		return
+	}
+	_ = fs.WriteFile(hashFile, []byte(hash), 0775)
+	return
+}
+
+func isPathForDirectory(fs filesystem.FS, dst string) bool {
+	if ok, _ := fs.IsDir(dst); ok {
+		return true
+	}
+	if filepath.Ext(dst) == "" {
+		return true
+	}
+	return strings.HasSuffix(dst, string(fs.PathSeparator())) || strings.HasSuffix(dst, "/")
+}
+
+// TransferFiles transfers a file from a location `src` to another `dest` and ensures the integrity (i.e. hash validation) of what was copied across.
+// `dest` can be a file or a directory. If not existent, it will be created on the fly. non-existent directory path should be terminated by a path separator i.e. / or \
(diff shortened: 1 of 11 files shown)


utils/sharedcache/common.go
```diff
@@ -0,0 +1,312 @@
...
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	if !fs.Exists(src) {
+		err = fmt.Errorf("%w: source path does not exist [%v]", commonerrors.ErrNotFound, src)
+		return
+	}
+	if result, suberr := fs.IsFile(src); !result || suberr != nil {
+		err = fmt.Errorf("%w: source is not a file [%v, (%v)]", commonerrors.ErrInvalid, src, suberr)
```
joshjennings98, line 80: subErr could be an issue with permissions (it calls stat) or something, not just that the source is not a file so I think that the error could be improved

utils/sharedcache/config.go
```diff
@@ -0,0 +1,32 @@
...
+	"time"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+
+	configUtils "github.com/ARM-software/golang-utils/utils/config"
+)
+
+type Configuration struct {
+	RemoteStoragePath       string        `mapstructure:"remote_storage_path"` // Path where the cache will be stored.
+	Timeout                 time.Duration `mapstructure:"Timeout"`             // Cache timeout if need be
```
joshjennings98, line 13: should be lowercase in the mapstructure?

utils/sharedcache/factory.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+const (
+	CacheMutable int = iota
+	CacheImmutable
+)
+
+var (
+	CacheTypes = []int{CacheMutable, CacheImmutable}
```
joshjennings98, line 16: where is this used

utils/sharedcache/factory.go
```diff
@@ -0,0 +1,27 @@
+package sharedcache
+
+import (
+	"fmt"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+const (
```
joshjennings98, line 10: I think it is better to do enums like:

```go
type Cache int

const (
	CacheMutable Cache = iota
	CacheImmutable
)
```
as it is more explicit and I think it means that you can't accidentally use a value that is out of range. You don't want people doing `NewCache(1, ....)` because if someone added `const x = 1` somewhere then `iota` will change the value of `CacheMutable` to avoid conflicts. Hence if we create the `type Cache` we are abstracting the enum away from being an int so people will have to go out of their way to use ints and they won't accidentally do something they shouldn't

utils/sharedcache/factory.go
```diff
@@ -0,0 +1,27 @@
...
+	CacheMutable int = iota
+	CacheImmutable
+)
+
+var (
+	CacheTypes = []int{CacheMutable, CacheImmutable}
+)
+
+func NewCache(cacheType int, fs filesystem.FS, cfg *Configuration) (ISharedCacheRepository, error) {
```
joshjennings98, line 19: if you do the change above then change the int to Cache

utils/sharedcache/sharedcache_immutable.go
```diff
@@ -0,0 +1,232 @@
...
+	return
+}
+
+func listCompleteFilesByModTime(ctx context.Context, fs filesystem.FS, entryDir string) (sorted []string, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+
+	// sort the files by mod time
```
joshjennings98, line 48: move this above the function declaration, it makes it seem like you are saying that Ls will sort by mod time

utils/sharedcache/sharedcache_immutable.go
```diff
@@ -0,0 +1,232 @@
...
+			fileModTime := statInfo.ModTime()
+			fileModTimes = append(fileModTimes, FileWithModTime{
+				filename: file,
+				modTime:  fileModTime,
+			})
+		}
+	}
+	// can't do this directly using strings and stattimes to check on the
+	// file because the sort.Slice requires the function to output bool
+	// so if we wouldn't be abele to return an error if stattimes failed
```
joshjennings98, line 78: able

utils/sharedcache/sharedcache_immutable.go
```diff
@@ -0,0 +1,232 @@
...
+	sort.Slice(fileModTimes, func(i, j int) bool { return fileModTimes[i].modTime.After(fileModTimes[j].modTime) })
+
+	// map to just string
+	for _, file := range fileModTimes {
```
joshjennings98, line 82: You told me a while ago that we should do:
```go
for i := range xxx {
   y = xxx[i]
   ...
}
```
Not:
```go
for _, y := range xxx {
   ...
}
```
Is this still the case? If so then this look should be changed to use the index

utils/sharedcache/sharedcache_mutable.go
```diff
@@ -0,0 +1,156 @@
...
+	// do the transfer
+	err = remoteLock.LockWithTimeout(ctx, s.lockTimeout)
+	if err != nil {
+		return
+	}
+	err = s.unpackPackageToLocalDestination(ctx, cachedPackage, dest)
+	if err != nil {
+		return
+	}
+	err = remoteLock.Unlock(ctx)
```
joshjennings98, line 78: this should be a defer between line 73 and 74 in case the unpack errors

utils/sharedcache/factory.go
acabarbaye: factory

utils/sharedcache/config.go
acabarbaye: good spot

utils/sharedcache/factory.go
acabarbaye: and in the tests

utils/sharedcache/sharedcache_immutable.go
acabarbaye: good spot

PR #122, opened 2022-08-23
⭐  `[filesystem]` get file ownership

acabarbaye:
### Description

`[filesystem]` get file ownership information.
This is to cover what is discussed here: https://github.com/golang/go/issues/14753

### Test Coverage

### DIFF
--- a/utils/filesystem/files.go
+++ b/utils/filesystem/files.go
@@ -68,11 +68,11 @@ func IdentityPathConverterFunc(path string) string {
 
 type VFS struct {
 	vfs           afero.Fs
-	fsType        int
+	fsType        FilesystemType
 	pathConverter func(path string) string
 }
 
-func NewVirtualFileSystem(vfs afero.Fs, fsType int, pathConverter func(path string) string) FS {
+func NewVirtualFileSystem(vfs afero.Fs, fsType FilesystemType, pathConverter func(path string) string) FS {
 	return &VFS{
 		vfs:           vfs,
 		fsType:        fsType,
@@ -84,7 +84,7 @@ func GetGlobalFileSystem() FS {
 	return globalFileSystem
 }
 
-func GetType() int {
+func GetType() FilesystemType {
 	return globalFileSystem.GetType()
 }
 
@@ -151,7 +151,7 @@ func (fs *VFS) walk(ctx context.Context, path string, info os.FileInfo, fn filep
 
 }
 
-func (fs *VFS) GetType() int {
+func (fs *VFS) GetType() FilesystemType {
 	return fs.fsType
 }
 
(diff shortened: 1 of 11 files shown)


utils/filesystem/files.go
```diff
@@ -695,6 +695,23 @@ func (fs *VFS) Chown(name string, uid, gid int) (err error) {
...
+	stat, err := fs.Stat(name)
+	if err != nil {
+		return
+	}
+	if stat == nil {
+		err = fmt.Errorf("%w: missing file info", commonerrors.ErrUndefined)
+		return
+	}
+	if reflection.IsEmpty(stat.Sys()) {
+		err = commonerrors.ErrNotImplemented
```
joshjennings98, line 708: ErrNotImplemented feels like the wrong error since .Sys() is calling the underlying OS stat function so that should never be not implemented? It feels like ErrUnexepected or ErrUndefined might be better, but this is a minor nitpick and it may be me misunderstanding .Sys() (i.e. you know it is being used somewhere where there it will be unavialable)

utils/filesystem/files_posix.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+)
+
+func determineFileOwners(info os.FileInfo) (uid, gid int, err error) {
+	if raw, ok := info.Sys().(*syscall.Stat_t); ok {
+		gid = int(raw.Gid)
+		uid = int(raw.Uid)
+	} else {
+		err = commonerrors.ErrNotImplemented
```
joshjennings98, line 18: same with above, in unix/linux at least stat will be implemented won't it? A failure would be another issue

utils/filesystem/filesystem.go
```diff
@@ -13,13 +13,15 @@ import (
 	"github.com/ARM-software/golang-utils/utils/commonerrors"
 )
 
+type FilesystemType int
```
joshjennings98, line 16: :+1:

joshjennings98 (dismissed):

utils/filesystem/files.go
acabarbaye: Sys() actually may not be implemented on some platforms or for some filesystems : For instance, https://github.com/spf13/afero/blob/2a70f2bb2db1524bf2aa3ca0cfebefa8d6367b7b/mem/file.go#L339. Hence, why this check is in place

utils/filesystem/files_posix.go
acabarbaye: So filesystems ay not define Sys() as `Stat_t` even if running on Linux and in this case, we want to say that we don't know what to do. I case `ErrNotSupported` may be better. I'll change this

acabarbaye: I change the error returned with some better description

joshjennings98 (approved):

PR #135, opened 2022-09-13
🔒 `[Unzip]` security risk mitigation

vana-01:
### Test Coverage


utils/filesystem/files_test.go
```diff
@@ -1306,6 +1306,37 @@ func TestFilepathStem(t *testing.T) {
...
+	testFile = "abovefilecountlimitzip"
+	srcPath = filepath.Join(testInDir, testFile+".zip")
+
+	destPath, err = fs.TempDirInTempDir("unzip-limits-")
+	assert.NoError(t, err)
+
+	defer fs.Rm(destPath)
+
+	_, err = fs.UnzipWithContextAndLimits(context.TODO(), srcPath, destPath, limits)
+	assert.NoError(t, err)
```
acabarbaye, line 1337: If the file is above the limit then an error should be raised
```suggestion
	assert.Error(t, err)
```
you should also decide on the type of error returned by doing a task like

The error returned should be of the type of one of the errors in this list https://github.com/ARM-software/golang-utils/blob/master/utils/commonerrors/errors.go

```
assert.True(t, commonerrors.Any(err, commonerrrors.ErrTooLarge) 
```

utils/filesystem/files_test.go
```diff
@@ -1306,6 +1306,37 @@ func TestFilepathStem(t *testing.T) {
 	})
 }
 
+func TestUnzipFileCountLimit(t *testing.T) {
+	fs := NewFs(StandardFS)
+
+	testInDir := "testdata"
```
acabarbaye, line 1312: for ease of readiness, you should create subtest so that the reporting gets clear
i.e.
```
t.Run("zip under the limits", func(){
....
})
```
then
```
t.Run("zip above the limits", func(){
....
})
```

utils/filesystem/files_test.go
```diff
@@ -1306,6 +1306,46 @@ func TestFilepathStem(t *testing.T) {
...
+		testFile := "abovefilecountlimitzip"
+		srcPath := filepath.Join(testInDir, testFile+".zip")
+
+		destPath, err := fs.TempDirInTempDir("unzip-limits-")
+		assert.NoError(t, err)
+		defer func() {
+			if tempErr := fs.Rm(destPath); tempErr != nil {
+					err = tempErr
+			}
+		}()
```
acabarbaye, line 1325: You can simplify it this way as you don't do anything with the error
```suggestion
		defer func() {
			_ := fs.Rm(destPath)
		}()
```

utils/filesystem/files.go
```diff
@@ -1223,8 +1223,10 @@ func (fs *VFS) unzip(ctx context.Context, source string, destination string, lim
 	if err != nil {
 		return
 	}
-	// List of file paths to return
 
+	var fileCounter int
```
acabarbaye, line 1227: I would suggest to use an atomic counter like `line 1230` so that we have some protection if we want later to parallelise any of the code

utils/filesystem/files.go
```diff
@@ -1259,6 +1261,8 @@ func (fs *VFS) unzip(ctx context.Context, source string, destination string, lim
 
 	// For each file in the zip file
 	for i := range zipReader.File {
+		fileCounter ++
```
acabarbaye, line 1264: see comment about the atomic

acabarbaye: if you use atomic, you can then call fileCounter.Inc()

utils/filesystem/files.go
```diff
@@ -1259,6 +1261,8 @@ func (fs *VFS) unzip(ctx context.Context, source string, destination string, lim
 
 	// For each file in the zip file
 	for i := range zipReader.File {
+		fileCounter.Inc()
```
acabarbaye, line 1264: 👍

acabarbaye (approved):

PR #175, opened 2022-12-15
✨ : [filesystem]  Allow `unzip` to be recursive for ease of use

joshjennings98:
### Description

Make unzip recursive for catching zip-bombs more effectively.

### Test Coverage


utils/filesystem/zip.go
```diff
@@ -223,8 +234,10 @@ func (fs *VFS) unzip(ctx context.Context, source string, destination string, lim
 			return fileList, subErr
 		}
 
-		// Keep list of files unzipped
-		fileList = append(fileList, filePath)
+		// Keep list of files unzipped (except zip files as they will be handled later)
+		if filepath.Ext(zippedFile.Name) != zipExt {
```
acabarbaye, line 238: the problem is that zip does not need to be with .zip, it could be .pack, .gz, .7zip ....

could you add a function `isZip(file string)` and use it here

PR #260, opened 2023-05-31
:sparkles: `[http]` Added `RetryOnError`

acabarbaye:
### Description

- Added `RetryOnError` in the `http` module to generically retry client calls on particular errors

### Test Coverage

### DIFF
--- a/utils/http/retry_test.go
+++ b/utils/http/retry_test.go
@@ -1,0 +1,94 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/commonerrors/errortest"
+	"github.com/ARM-software/golang-utils/utils/logs/logstest"
+)
+
+func TestRetryOnError(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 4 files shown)


utils/http/retry_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/commonerrors/errortest"
+	"github.com/ARM-software/golang-utils/utils/logs/logstest"
+)
+
+func TestRetryOnError(t *testing.T) {
+	tests := []struct {
+		policy *RetryPolicyConfiguration
+	}{
+		{
```
joshjennings98, line 21: can we test nil? so that if the implementation is changed in the future and someone forgets to account for nil, we can catch it

utils/http/retry_test.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+		{
+			policy: DefaultNoRetryPolicyConfiguration(),
+		},
+		{
+			policy: DefaultLinearBackoffRetryPolicyConfiguration(),
+		},
+		{
+			policy: DefaultExponentialBackoffRetryPolicyConfiguration(),
+		},
```
joshjennings98, line 32: also I know it is annoying, but I think we should have a cancelled context test. The `.Context(ctx)` should handle this, but in the event we switch implementations one day it would be good to have a test to make sure we account for cancelled contexts. Especially since cancelling http contexts seems like it is a common thing that might occur

joshjennings98 (approved):

PR #274, opened 2023-06-09
:sparkles: `[filesystem]` Created Adapter to `embed.FS` 

acabarbaye:
### Description
- `[filesystem]` Created Adapter to `embed.FS` for use of common utilities"
- `[filesystem]` Ability to read the content of an `afero.File`

### Test Coverage

### DIFF
--- a/utils/filesystem/embedfs_test.go
+++ b/utils/filesystem/embedfs_test.go
@@ -1,0 +1,154 @@
+package filesystem
+
+import (
+	"embed"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"testing"
+
+	"github.com/bxcodec/faker/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/commonerrors/errortest"
+)
+
+//go:embed *
+var testContent embed.FS
+
+const testFile1Content = "this is a text file with some content\n"
+
+func Test_embedFS_Exists(t *testing.T) {
+	fs, err := NewEmbedFileSystem(&testContent)
+	require.NoError(t, err)
+	assert.False(t, fs.Exists(faker.DomainName()))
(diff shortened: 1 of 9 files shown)


utils/filesystem/embedfs_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/bxcodec/faker/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/commonerrors/errortest"
+)
+
+//go:embed *
```
joshjennings98, line 19: should this embed the files we are testing. I don't know how `*` will behave?

utils/filesystem/filesystem.go
```diff
@@ -42,12 +52,12 @@ func NewFs(fsType FilesystemType) FS {
 	return NewStandardFileSystem()
 }
 
-// Converts file system error into common errors
-func ConvertFileSytemError(err error) error {
+// ConvertFileSystemError converts file system error into common errors
```
joshjennings98, line 55: I know this isn't necessarily from this change, but I think for discovery reasons this should be in commonerrors. That is where I assume all the convert to common errors stuff it

joshjennings98 (approved):
Only two comments but they are nitpicks

utils/filesystem/embedfs_test.go
acabarbaye: this embeds all files in current directory only when building the test binaries

utils/filesystem/filesystem.go
acabarbaye: each module has its own convertError function because the errors are specific to their domain. Here to do with files systems, other places to do with hashing, etc.. If we put them in the commonerrors, it means everytime the commonerror module is loader (almost everywhere) all the errors and modules imported in that module would also be loaded. meaning every user of common loaded would load os, hash, ... even if not used. Having them in the module they are related to means there is a separation of concern. The fact that the context error are converted in the commonerrors is only because context is pretty much used everywhere too but we could have created a context module

PR #290, opened 2023-07-24
:sparkles: Added `[logrimp]` for typical logr.Logger implementations

acabarbaye:
### Description

This module provides a few implementations of the `logr.Logger` interface to cover most typical usecases and ease their use and initialisation.

### Test Coverage


utils/logs/logrimp/logr_test.go
```diff
@@ -0,0 +1,54 @@
...
+			Logger: NewLogrusLogger(logrus.New()),
+			name:   "Logrus",
+		},
+	}
+	for i := range tests {
+		test := tests[i]
+		t.Run(test.name, func(t *testing.T) {
+			logger := test.Logger
+			logger.WithName(faker.Name()).WithValues("foo", "bar").Info(faker.Sentence())
+			logger.Error(commonerrors.ErrUnexpected, faker.Sentence(), faker.Word(), faker.Name())
```
Luke-Parkin, line 49: How do you check the logger actually worked here?

Luke-Parkin (approved):
lgtm, just some questions

utils/logs/logrimp/logr_test.go
acabarbaye: this is using most `logr` interface functions. If there is any issue with the implementation, there will be either a compilation error or a go panic. In both cases, this would make the test to fail. The goal of these tests is to ensure the versions of the interface and implementation are compatible and that we can safely use all the typical functions, not the actual implementation logic if that makes sense

utils/logs/std_logger.go
```diff
@@ -76,7 +76,7 @@ type stdAdaptor struct {
 	loggers Loggers
 }
 
-func (s *stdAdaptor) Output(calldepth int, logline string) error {
+func (s *stdAdaptor) Output(_ int, logline string) error {
```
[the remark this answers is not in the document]
acabarbaye: no this is because the parameter is actually not used. linter complains when this is the case

utils/logs/logrimp/logr_test.go
Luke-Parkin: That does, thanks.

utils/logs/std_logger.go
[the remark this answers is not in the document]
acabarbaye: it is to fit the interface the adaptor implements

PR #292, opened 2023-07-28
:sparkles: `[filesystem]` Support for `ZipFS`

acabarbaye:
### Description

- Added support for  `ZipFS`

### Test Coverage


utils/filesystem/zipfs_test.go
```diff
@@ -0,0 +1,264 @@
...
+	assert.True(t, fs.Exists("testDir1"))
+	assert.True(t, fs.Exists("testFile"))
+	assert.True(t, fs.Exists("testDir1/testFile"))
+	require.NoError(t, zipFile.Close())
+}
+
+func Test_zipFS_FileInfo(t *testing.T) {
+	fs, zipFile, err := NewZipFileSystemFromStandardFileSystem(filepath.Join("testdata", "testunzip.zip"), NoLimits())
+	require.NoError(t, err)
+	defer func() {
```
joshjennings98, line 62: Since this is a bit ugly and we would have to do it any time we used this new function, could we maybe have a closeZip() function that we can more easily defer

utils/filesystem/zipfs_test.go
```diff
@@ -0,0 +1,264 @@
...
+func Test_zipFs_Browsing2(t *testing.T) {
+	zipFs, zipFile, err := NewZipFileSystemFromStandardFileSystem(filepath.Join("testdata", "t.zip"), NoLimits())
+	require.NoError(t, err)
+	defer func() {
+		if zipFile != nil {
+			_ = zipFile.Close()
+		}
+	}()
+	require.NotNil(t, zipFile)
+	var wFunc = func(path string, info fs.FileInfo, err error) error {
```
joshjennings98, line 116: could we pre generate a list of the files using ListDir and then walk it and make sure they match instead of printing

utils/filesystem/zipfs_test.go
joshjennings98, line 1: could the names of the .Run functions be better than:

xxx
xxx 2

just so it is clear what the different methods are

joshjennings98: and the same with itemType itemType2

joshjennings98 (approved):

utils/filesystem/zipfs_test.go
acabarbaye: good idea. will add it

utils/filesystem/zipfs_test.go
acabarbaye: will do

utils/filesystem/zipfs_test.go
acabarbaye: introduced a CloseableFilesystem for this

joshjennings98 (approved):
that's nicer

PR #296, opened 2023-08-01
:sparkles: Added module `[encryption]` for simple cryptography utilities

acabarbaye:
### Description

- Added module `[encryption]` for simple cryptography utilities
- Support more hashing algorithm

### Test Coverage


joshjennings98 (dismissed):
I think it's okay, I can't see anything that makes me think there will be a problem

utils/encryption/encryption.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+const KeySize = 32
+
+var (
+	keySizeError = fmt.Errorf("%w: recipient key has invalid size (expected %d bytes)", commonerrors.ErrInvalid, KeySize)
+)
+
+// GenerateKeyPair generates a asymmetric key pair suitable for use with encryption utilities. Works with [NaCl box](https://nacl.cr.yp.to/box.html.)
+func GenerateKeyPair() (base64EncodedPublicKey, base64EncodedPrivateKey string, err error) {
```
joshjennings98, line 20: Maybe these should be a more complex type than a string. Maybe make it harder to read the keys by having them in a mutex or something and make getters compulsory so you can't just accidentally log the string containing the key.

I am not sure

acabarbaye: made it a more complex type

utils/encryption/encryption.go
```diff
@@ -0,0 +1,126 @@
...
+
+const KeySize = 32
+
+var (
+	errKeySize = fmt.Errorf("%w: recipient key has invalid size", commonerrors.ErrInvalid)
+)
+
+type keyPair struct {
+	public  string
+	private string
```
joshjennings98, line 21: I know you shouldn't use pointers to primitives but would it be worth having these (at least the private key) as pointers so that doing `Log`/`Printf`/`Println` doesn't expose the key? Even though they aren't exposed and the `keyPair` isn't exposed, you can still print the resulting `IKeyPair` and see the contents. Wity pointers you won't get the values themselves.

I know it doesn't completely solve the problem as you can use the "unsafe" module and get the value at the pointer, but it might help prevent leaking information in logs. 

Whether it's worth the effort to do this, is up to you.

joshjennings98 (dismissed):
I think this is better now

joshjennings98 (approved):

PR #313, opened 2023-09-15
:sparkles: `[platform]` Add platform user management

acabarbaye:
### Description

- Adding utilities with regards to platform user management

### Test Coverage

- [ ]  This change is covered by existing or additional automated tests.
- [x]  Additional tests are not required for this change (e.g. documentation update).


utils/platform/users.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+	found, _ := HasUser(username)
+	if found {
+		return nil
+	}
+	return ConvertUserGroupError(addUser(ctx, username, fullname, password))
+}
+
+// DeleteUser removes a user from the platform
```
joshjennings98, line 39: Can there be more of a distinction between when to use remove Vs delete user? If I was importing this library I wouldn't know which to use

joshjennings98 (dismissed):
Just change the descriptions around delete and remove

utils/platform/users.go
acabarbaye, line 39:
```suggestion
// DeleteUser removes a user from the platform when the user is specified using a `user.User` structure.
```

acabarbaye: tried to do so

PR #341, opened 2023-10-17
:sparkles: `[filesystem]` Allow Force deletion

acabarbaye:
### Description
- `[filesystem]` Sometimes deletion of path is not possible because they are used by a different process, etc.
This adds a function for forcing the deletion by taking escalated privileges (superuser permissions) to carry out the action

### Test Coverage


joshjennings98 (dismissed):
thanks for doing this

PR #357, opened 2023-11-01
:sparkles: [`filesystem`] Add various utilities

acabarbaye:
### Description

- Add Glob search
- Add `Touch`
- Add recursive utilities

### Test Coverage


utils/filesystem/files_test.go
```diff
@@ -133,6 +130,64 @@ func TestOpen(t *testing.T) {
...
+			assert.True(t, fs.Exists(tmpFile))
+			empty, _ := fs.IsEmpty(tmpFile)
+			assert.True(t, empty)
+			timesFile1, err := fs.StatTimes(tmpFile)
+			require.NoError(t, err)
+
+			tmpFile2 := filepath.Join(tmpDir, "testtouch2.test")
+			err = fs.Touch(tmpFile2)
+			require.NoError(t, err)
+			empty, _ = fs.IsEmpty(tmpFile2)
```
joshjennings98, line 152: I'd probably keep the error and require it as noerror. For example, If touch had caused a permission change you would get an error and empty would be false but the false would be due to an error not because its empty etc.

joshjennings98 (approved):
Looks good, just one comment

utils/filesystem/files_test.go
acabarbaye: fair enough

PR #443, opened 2024-04-03
:sparkles: Improve process termination to avoid orphan subprocesses

acabarbaye:
### Description

- As described in https://github.com/golang/go/issues/24050, processes could be left hanging if not terminated properly. Better process termination was introduced to avoid this phenomenon.
- IsProcessRunning was added to `/proc` to check if a process identified by its pid is currently running.

### Test Coverage

### DIFF
--- a/changes/20240403145328.bugfix
+++ b/changes/20240403145328.bugfix
@@ -1,0 +1,1 @@
+:bug: Fix hanging sub processes as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration
--- a/changes/20240403145328.bugfix
+++ b/changes/20240403145328.bugfix
@@ -1,2 +1,1 @@
 :bug: Fix hanging sub processes when deleting folders as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration
-
--- a/changes/20240403145328.bugfix
+++ b/changes/20240403145328.bugfix
@@ -1,1 +1,2 @@
-:bug: Fix hanging sub processes as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration
+:bug: Fix hanging sub processes when deleting folders as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration
+
--- a/utils/proc/process.go
+++ b/utils/proc/process.go
@@ -53,27 +53,30 @@ func FindProcess(ctx context.Context, pid int) (p IProcess, err error) {
 	return
 }
 
+// IsProcessRunning states whether a process is running or not. An error is returned if the context is Done while looking for the process state.
+func IsProcessRunning(ctx context.Context, pid int) (running bool, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	p, SubErr := FindProcess(ctx, pid)
+	switch {
+	case SubErr == nil:
+		running = p.IsRunning()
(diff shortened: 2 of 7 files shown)


changes/20240403145328.bugfix
```diff
@@ -0,0 +1 @@
+:bug: Fix hanging sub processes as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration
```
acabarbaye, line 1:
```suggestion
:bug: Fix hanging sub processes when deleting folders as described in [(go issue)](https://github.com/golang/go/issues/24050). See https://github.com/go-cmd/cmd/tree/master?tab=readme-ov-file#proper-process-termination for inspiration

```

utils/proc/process.go
```diff
@@ -53,27 +53,30 @@ func FindProcess(ctx context.Context, pid int) (p IProcess, err error) {
...
+// IsProcessRunning states whether a process is running or not. An error is returned if the context is Done while looking for the process state.
+func IsProcessRunning(ctx context.Context, pid int) (running bool, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	p, SubErr := FindProcess(ctx, pid)
+	switch {
+	case SubErr == nil:
+		running = p.IsRunning()
```
joshjennings98, line 65: based on L223, does this return an error? Or should the one on L223 not return an error?

Or is it a different process type?

utils/subprocess/executor_test.go
```diff
@@ -431,3 +436,47 @@ func TestCancelledSubprocess3(t *testing.T) {
...
+	logger, err := logs.NewLogrLogger(logstest.NewTestLogger(t), "test")
+	require.NoError(t, err)
+	t.Run("happy", func(t *testing.T) {
+		output, err := OutputWithEnvironment(context.Background(), logger, nil, "du", "-h")
+		require.NoError(t, err)
+		assert.NotEmpty(t, output)
+		fmt.Println(output)
+	})
+	t.Run("happy with output", func(t *testing.T) {
+		testString := fmt.Sprintf("This is a test %v!", faker.Sentence())
```
joshjennings98, line 454: You night want to wrap these in `'` just in case `faker.Sentence()` creates some valid shell code (as unlikely as that is...)
```suggestion
		testString := fmt.Sprintf("'This is a test %v!'", faker.Sentence())
```

Same for the other ones

joshjennings98 (approved):
I just had a couple of comments but it looks good :+1:

utils/proc/process.go
acabarbaye: I tried to describe it in the function comment but the error is only to describe whether the context was cancelled or timed out

utils/proc/process.go
```diff
@@ -209,6 +216,29 @@ func killChildren(ctx context.Context, p *process.Process) (err error) {
...
+	}
+	// from man 2 kill: If sig is 0, then no signal is sent, but error checking is still performed; this can be used to check for the existence of a process ID or process group ID.
+	exist, _ := process.PidExists(p.Pid)
+	if !exist {
+		running = false
+		return
+	}
+	status, err := p.Status()
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
acabarbaye: so this is because, we are trying to be more granular by checking the status. Nonetheless, on some platform such as Windows, the status is not implemented and so in this case we want to fallback to the IsRunning method, which is not thorough on unix platform as a zombie process is reported as running

utils/proc/process.go
```diff
@@ -209,6 +216,29 @@ func killChildren(ctx context.Context, p *process.Process) (err error) {
...
+	if !running {
+		return
+	}
+	// from man 2 kill: If sig is 0, then no signal is sent, but error checking is still performed; this can be used to check for the existence of a process ID or process group ID.
+	exist, _ := process.PidExists(p.Pid)
+	if !exist {
+		running = false
+		return
+	}
+	status, err := p.Status()
```
acabarbaye, line 233:
```suggestion
        // On some platforms, such as *nix, a zombie process is reported as a running process by p.IsRunning() but this is not the case. Therefore, a further check is performed on the process status to verify a running process is actually in the expected running state. Nonetheless, status is not cross platform and is not implemented on Windows. For those platform, the status returned by IsRunning is then considered
	status, err := p.Status()
```

PR #476, opened 2024-06-13
:sparkles: `[logs]` Add a way to use a `Loggers` as a plain `logr.Logger`

acabarbaye:
### Description

- added a new logger adaptor

### Test Coverage


joshjennings98 (approved):
I think this all looks good. The only very minor nitpick is that there is a fair amount of nested if statements but I don't think it really matters

PR #497, opened 2024-08-28
✨ `[filesystem]` Added LsRecursive in `filesystem` to browse file system trees

ERICwangyiquan:
### Description

- Introduced a way to perform recursive `ls`

### Test Coverage


acabarbaye:
could you add a small description to this PR

changes/20240828133012.feature
```diff
@@ -0,0 +1 @@
+✨ Added LsRecursive
```
joshjennings98, line 1: expand on this, explain what it is for (and how it is different from the existing functions) but keep it one line

acabarbaye, line 1:
```suggestion
✨ `[filesystem]` Added `LsRecursive` to browse file system trees and list their content
```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
 	return
 }
 
+func (fs *VFS) LsRecursive(ctx context.Context, dir string) (files []string, err error) {
+	limits := NoLimits()
+	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, limits)
```
joshjennings98, line 1133: just do this inline
```suggestion
	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, NoLimits())
```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+}
+
+func (fs *VFS) LsRecursiveWithExtensionPatterns(ctx context.Context, dir string, exclusionPatterns ...string) (files []string, err error) {
+	limits := NoLimits()
+	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, limits, exclusionPatterns...)
```
joshjennings98, line 1138: ditto
```suggestion
	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, NoLimits(), exclusionPatterns...)
```

utils/filesystem/files.go
acabarbaye, line 1133:
```suggestion
	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, NoLimits())
```

joshjennings98: I think it only makes sense to instantiate a lone variable if it is being used in multiple places

utils/filesystem/files.go
acabarbaye, line 1138:
```suggestion
	return fs.LsRecursiveWithExtensionPatternsAndLimits(ctx, dir, NoLimits(), exclusionPatterns...)
```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+}
+
+func (fs *VFS) LsRecursiveWithExtensionPatternsAndLimits(ctx context.Context, dir string, limits ILimits, exclusionPatterns ...string) (files []string, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	if limits == nil {
+		err = fmt.Errorf("%w: missing file system limits definition", commonerrors.ErrUndefined)
```
joshjennings98, line 1147: it isn't a definition
```suggestion
		err = fmt.Errorf("%w: missing file system limits", commonerrors.ErrUndefined)
```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+		// In this case, the whole tree is excluded and so, we can stop here
+		return
+	}
+
+	fn := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		relativePath := strings.TrimPrefix(path, dir)
```
acabarbaye, line 1164: use filepath.Rel instead as `path` and `dir` may use different path separators

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+			return filepath.SkipDir
+		}
+
+		if limits.Apply() && int64(len(files)) >= limits.GetMaxFileCount() {
+			return filepath.SkipAll
+		}
+
+		if !info.IsDir() {
+			files = append(files, path)
+		}
```
joshjennings98, line 1177:
```suggestion
		}

```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+		return
+	}
+
+	fn := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		relativePath := strings.TrimPrefix(path, dir)
+		currentDepth := int64(strings.Count(relativePath, string(fs.PathSeparator())))
```
acabarbaye, line 1165: use the already defined function for this https://github.com/ARM-software/golang-utils/blob/427207db9c482d6bfe81ede28eb30c5338665ed6/utils/filesystem/filepath.go#L22

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+
+		if limits.Apply() && currentDepth > limits.GetMaxDepth() {
+			return filepath.SkipDir
+		}
+
+		if limits.Apply() && int64(len(files)) >= limits.GetMaxFileCount() {
+			return filepath.SkipAll
```
joshjennings98, line 1172: If we look at https://github.com/ARM-software/golang-utils/blob/427207db9c482d6bfe81ede28eb30c5338665ed6/utils/filesystem/files.go#L143 then we can see that if you do filepath.SkipAll then it will return the error err=filepath.SkipAll from the parent function since it isn't handled in the same way as SkipDir.

We want to make sure that any errors that are returned from golang-utils are wrapped in our commonerrors so this one should probably return something like `commonerrors.ErrTooLarge` and wrapped with a nice error message saying the number of files exceeds the limit

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,60 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+		}
+
+		if !info.IsDir() {
```
acabarbaye, line 1175: this goes against the description in the interface: `list all files and directories`. Either remove this check or alter the documentation of the function or add an option to the function

utils/filesystem/files.go
joshjennings98, line 1165: this might not work if you escape the path separator, e.g.:


`/dir1/dir2` == depth 2
`/dir1\/escaped` == depth 1 but by just counting the separators you will have depth 2 which is wrong

joshjennings98: You should turn it into a path object instead of doing string operations

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+		filepath.Join(rootDir, "dir2", "dir5", "test5.jar"),
+	}
+
+	for _, dir := range dirs {
+		if err := vfs.MkdirAll(dir, 0755); err != nil {
+			return err
+		}
+	}
+	for _, file := range files {
+		err := afero.WriteFile(vfs, file, []byte("test content"), 0644)
```
acabarbaye, line 2053: use vfs.Touch

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
 	require.NoError(t, osFile.Close())
 	require.Error(t, rFile.Close(), "file must already be closed")
 }
+
+func createTestFileStructure(vfs afero.Fs, rootDir string) error {
```
acabarbaye, line 2034:
```suggestion
func generateTestTree(vfs afero.Fs, rootDir string) error {
```

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+	}
+	return nil
+}
+
+func TestVFS_LsRecursive(t *testing.T) {
+	fs := globalFileSystem
+
+	vfs := afero.NewOsFs()
+
+	rootDir, err := afero.TempDir(vfs, "", "testdata")
```
acabarbaye, line 2066: do not use the afero library but rahte rour filesystem function

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+	}
+	tests := []struct {
+		name           string
+		f              func(tree string) ([]string, error)
+		fields         fields
+		args           args
+		wantFiles      []string
+		wantErr        assert.ErrorAssertionFunc
+		wantFilesCount int
+	}{
```
joshjennings98, line 2091: to add to my comments, add some tests with escaped file path separators

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+
+func TestVFS_LsRecursive(t *testing.T) {
+	fs := globalFileSystem
+
+	vfs := afero.NewOsFs()
+
+	rootDir, err := afero.TempDir(vfs, "", "testdata")
+	if err != nil {
+		t.Fatalf("Failed to create temp directory: %v", err)
+	}
```
acabarbaye, line 2069: use require.NoError(t, ..) instead

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+	defer func() { _ = vfs.RemoveAll(rootDir) }()
```
joshjennings98, line 2070: you can just do `rootDir := t.TempDir()` and not worry about the cleanup or errors.

acabarbaye:
do not forget to run `go generate ./...` so that all the mocks are populated

utils/filesystem/interfaces.go
```diff
@@ -200,10 +200,18 @@ type FS interface {
...
+	LsRecursive(ctx context.Context, dir string) (files []string, err error)
 	// LsWithExclusionPatterns lists all files and directory (equivalent to ls) but exclude the ones matching the exclusion patterns.
 	LsWithExclusionPatterns(dir string, exclusionPatterns ...string) (files []string, err error)
+	// LsRecursiveWithExtensionPatterns lists all files and directory recursively, including subdirectories but exclude the ones matching the exclusion patterns.
+	LsRecursiveWithExtensionPatterns(ctx context.Context, dir string, exclusionPatterns ...string) (files []string, err error)
+	// LsRecursiveWithExtensionPatternsAndLimits lists all files and directory recursively, including subdirectories but exclude the ones matching the exclusion patterns and add some limits for recursion
+	LsRecursiveWithExtensionPatternsAndLimits(ctx context.Context, dir string, limit ILimits, exclusionPatterns ...string) (files []string, err error)
 	// LsFromOpenedDirectory lists all files and directory (equivalent to ls)
 	LsFromOpenedDirectory(dir File) (files []string, err error)
+	// LsRecursiveFromOpenedDirectory lists all files and directory recursively
```
joshjennings98, line 213: your code is not including directories in it's output so this description is wrong is wrong:

```
	if !info.IsDir() {
			files = append(files, path)
		}
```

utils/filesystem/files_test.go
joshjennings98, line 2034: you should use the filesystem representations from this module instead of afero.Fs

Although I don't think you need to use a different filesystem at all, just use GetGlobalFileSystem() since you are writing to the filesystem.

This might make more sense if you were testing the in memory file system as well

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+
+	if err := createTestFileStructure(vfs, rootDir); err != nil {
+		t.Fatalf("Failed to create test file structure: %v", err)
+	}
+
+	type fields struct {
+		vfs afero.Fs
+	}
```
joshjennings98, line 2078: what are you using this for?

utils/filesystem/files_test.go
joshjennings98, line 1: I think you should approach the tests in a different way.

If you look at how we do the tests for the other functions I think that they are a bit clearer than the table driven tests that Golang generated for you.

For example https://github.com/ARM-software/golang-utils/blob/427207db9c482d6bfe81ede28eb30c5338665ed6/utils/filesystem/files_test.go#L930 Here we have some helpers that generate the file tree (although it doesn't use t.TmpDir...) and it tests both file system types. It does this without complicating the tests with a bunch of structs.

I think the table driven tests are good but I think you should simplify them as the ones generated by Goland are a bit complex and harder to follow. This will also keep the tests more consistent with the other tests

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,261 @@ func TestConvertToOSFile(t *testing.T) {
...
+			wantFiles: []string{
+				filepath.Join(rootDir, "dir2", "dir5", "test4.txt"),
+				filepath.Join(rootDir, "dir2", "dir5", "test5.jar"),
+			},
+			wantErr: func(t assert.TestingT, err error, msgAndArgs ...interface{}) bool {
+				if errors.Is(err, filepath.SkipDir) {
+					return true
+				}
+				return assert.NoError(t, err, msgAndArgs...)
+			},
```
joshjennings98, line 2231: having functions returned like this makes it hard to follow when reading the test as you have to add conditional logic to parse the response

joshjennings98:
You have made a really good start here. The function itself looks good, I just think the tests could be improved. Have a look at https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/. It provides an overview of AAA tests which keep the tests simple and easy to follow. We can't always keep tests simple but it is better if we do

changes/20240828133012.feature
```diff
@@ -0,0 +1 @@
+✨  `[filesystem]` Added `LsRecursive` to browse file system trees and list their content
```
joshjennings98, line 1: This is much better, you should do it like this for all the newsfiles :+1:

utils/filesystem/gomock_reflect_1311679817/prog.go
joshjennings98, line 1: I don't know what this is for, I think it might have been generated by mistake?

utils/filesystem/files.go
```diff
@@ -1160,6 +1205,23 @@ func (fs *VFS) LsFromOpenedDirectory(dir File) ([]string, error) {
...
+	if err != nil {
+		return
+	}
+	err = fs.checkWhetherUnderlyingResourceIsClosed()
+	if err != nil {
+		return nil, fmt.Errorf("%w: underlying resource is closed", commonerrors.ErrUndefined)
+	}
+	if dir == nil {
+		return nil, fmt.Errorf("%w: nil directory", commonerrors.ErrUndefined)
+	}
```
joshjennings98, line 1219: Move this to just after the DetermineContextError part.

This is like a check that the arguments are valid so it should be before the business logic. I think the code checking the arguments should be kept together.

Also I think the error description should not mention nil, say something like fmt.Errorf("%w: supplied directory was undefined", commonerrors.ErrUndefined)

joshjennings98:
This is good, I just had a few comments but once they are done I think we can merge it :+1:

utils/filesystem/interfaces.go
```diff
@@ -200,10 +200,18 @@ type FS interface {
...
 	Ls(dir string) (files []string, err error)
+	// LsRecursive lists all files recursively, including subdirectories
+	LsRecursive(ctx context.Context, dir string) (files []string, err error)
 	// LsWithExclusionPatterns lists all files and directory (equivalent to ls) but exclude the ones matching the exclusion patterns.
 	LsWithExclusionPatterns(dir string, exclusionPatterns ...string) (files []string, err error)
+	// LsRecursiveWithExtensionPatterns lists all files recursively, including subdirectories but exclude the ones matching the exclusion patterns.
+	LsRecursiveWithExtensionPatterns(ctx context.Context, dir string, exclusionPatterns ...string) (files []string, err error)
+	// LsRecursiveWithExtensionPatternsAndLimits lists all files recursively, including subdirectories but exclude the ones matching the exclusion patterns and add some limits for recursion
+	LsRecursiveWithExtensionPatternsAndLimits(ctx context.Context, dir string, limit ILimits, exclusionPatterns ...string) (files []string, err error)
 	// LsFromOpenedDirectory lists all files and directory (equivalent to ls)
```
acabarbaye, line 211:
```suggestion
	// LsFromOpenedDirectory lists all files and directories (equivalent to ls)
```

utils/filesystem/files.go
```diff
@@ -1129,6 +1128,53 @@ func (fs *VFS) LsWithExclusionPatterns(dir string, exclusionPatterns ...string)
...
+
+		if limits.Apply() && currentDepth >= limits.GetMaxDepth() {
+			return filepath.SkipDir
+		}
+
+		if limits.Apply() && int64(len(files)) >= limits.GetMaxFileCount() {
+			return fmt.Errorf("number of files exceeds the limit of %d: %w", limits.GetMaxFileCount(), commonerrors.ErrTooLarge)
+		}
+
+		if !info.IsDir() {
```
acabarbaye, line 1167: this condition check will have to be removed since the description of the interface states `list all files recursively, including subdirectories`

acabarbaye: I would suggest a parameter flag to be added to the function signature such as `includeDirectories bool`

utils/go.mod
```diff
@@ -93,6 +93,7 @@ require (
 	github.com/tklauser/numcpus v0.6.1 // indirect
 	github.com/xanzy/ssh-agent v0.3.3 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
+	go.uber.org/mock v0.4.0 // indirect
```
acabarbaye, line 96: I don't think this dependency is necessary. I believe it may have been brought up when you regenerate the mocks. Could you keep the golang standard library implementation?

utils/filesystem/gomock_reflect_1311679817/prog.go
acabarbaye: yes please remove this file. These are temporary files generated during mock generation and there should not be committed in

utils/filesystem/files_test.go
```diff
@@ -2030,3 +2030,314 @@ func TestConvertToOSFile(t *testing.T) {
 	require.NoError(t, osFile.Close())
 	require.Error(t, rFile.Close(), "file must already be closed")
 }
+
+func TestLsRecursive(t *testing.T) {
```
acabarbaye, line 2034: really good!

acabarbaye (approved):

joshjennings98 (approved):
This is good, go ahead and merge it :+1:

PR #549, opened 2025-01-13
:sparkles: `[collections]` Added `Conditions` type to provide a helper to deal with flags or boolean slices

acabarbaye:
### Description

- helpers for evaluating flags

### Test Coverage

### DIFF
--- a/utils/collection/conditions.go
+++ b/utils/collection/conditions.go
@@ -1,0 +1,213 @@
+package collection
+
+import (
+	"fmt"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+)
+
+type Conditions []bool
+
+// NewConditions creates a set of conditions.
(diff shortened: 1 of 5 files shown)


utils/collection/conditions.go
```diff
@@ -0,0 +1,213 @@
...
+
+// AllTrue returns whether all values are true.
+func AllTrue(values ...bool) bool {
+	return All(values)
+}
+
+// Negate returns the slice with contrary values.
+func Negate(values ...bool) []bool {
+	if len(values) == 0 {
+		return nil
```
joshjennings98, line 162: i think that a nil slice is technically different from an empty slice so if the slice is empty but not nil then it should return `[]bool{}` or `make([]bool, 0)

acabarbaye: I addressed this

joshjennings98 (approved):

PR #562, opened 2025-02-11
:sparkles: `encryption` Add support for encrypting content using hybrid RSA/AES encryption

joshjennings98:
### Description

This change provides a method of encrypting a payload using hybrid RSA AES encryption. The payload itself is encrypted with the AES key which is fast but symmetric meaning both parties require the key. Therefore the key is encrypted with asymmetric RSA encryption and the encrypted key is included in the payload itself. RSA is not efficient on larger payloads so this method allows for fast encryption whilst also being asymmetric.

Information on hybrid RSA/AES encryption can be found here https://www.ijrar.org/papers/IJRAR23B1852.pdf

### Test Coverage

### DIFF
--- a/utils/encryption/aesrsa/encryption.go
+++ b/utils/encryption/aesrsa/encryption.go
@@ -41,10 +41,10 @@ func ParsePEMBlock(path string) (block *pem.Block, err error) {
 	return
 }
 
-// DecryptHybridAESRSAEncryptedPayload takes a path to an RSA private key and uses it to decode the AES key in a
-// hybrid encoded payload. This AES key is then used to decode the actual payload contents. Information of the use
-// of hybrid AES RSA encryption can be found here https://www.ijrar.org/papers/IJRAR23B1852.pdf
-func DecryptHybridAESRSAEncryptedPayload(privateKeyPath string, payload *HybridAESRSAEncryptedPayload) (decrypted []byte, err error) {
+// DecryptHybridAESRSAEncryptedPayloadFromPrivateKey takes a path to an RSA private key and uses it to decode the AES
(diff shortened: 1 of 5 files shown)


utils/encryption/aesrsa/encryption.go
```diff
@@ -0,0 +1,162 @@
...
+		err = fmt.Errorf("%w: could not parse PEM block from '%v': %v", commonerrors.ErrUnexpected, privateKeyPath, err.Error())
+		return
+	}
+
+	if block == nil {
+		err = fmt.Errorf("%w: block was empty", commonerrors.ErrEmpty)
+		return
+	}
+
+	priv, err := x509.ParsePKCS1PrivateKey(block.Bytes)
```
acabarbaye, line 59: could we have the following in its own function to work on a byte array

utils/encryption/aesrsa/payload.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return validation.ValidateStruct(p,
+		validation.Field(&p.CipherText, validation.Required),
+		validation.Field(&p.EncryptedKey, validation.Required),
+		validation.Field(&p.Nonce, validation.Required),
+	)
+}
+
+func DecodeHybridAESRSAEncryptedPayload(p *HybridAESRSAEncryptedPayload) (cipher, key, nonce []byte, err error) {
+	key, err = base64.StdEncoding.DecodeString(p.EncryptedKey)
```
acabarbaye, line 36: check if p is not nil

acabarbaye (approved):

acabarbaye (approved):

PR #569, opened 2025-02-27
:sparkles: `[semver]` Utilities to determine the version of a project from a version file generated by autoversion

acabarbaye:
### Description

- Utilities to determine the version of a project from a version file generated by [autoversion](https://github.com/ARMmbed/autoversion)

### Test Coverage


utils/filesystem/embedfs.go
```diff
@@ -11,7 +10,7 @@ import (
 
 func newEmbedFSAdapter(fs *embed.FS) (afero.Fs, error) {
 	if fs == nil {
-		return nil, fmt.Errorf("%w: missing filesystem", commonerrors.ErrUndefined)
+		return nil, commonerrors.UndefinedVariable("embedded file system", "")
```
joshjennings98, line 13: Would it be better to have two functions so we don't clutter it with the version that has an empty "" string arg

utils/semver/version.go
```diff
@@ -0,0 +1,69 @@
...
+
+// DetermineVersionFromFile retrieves the information about the version of a project based on a version file generated by [autoversion](https://github.com/ARMmbed/autoversion).
+// The Canonical version is returned.
+func DetermineVersionFromFile(ctx context.Context, fs filesystem.FS, versionFilePath string) (version string, err error) {
+	err = parallelisation.DetermineContextError(ctx)
+	if err != nil {
+		return
+	}
+	if fs == nil {
+		err = commonerrors.UndefinedVariable("file system", "")
```
joshjennings98, line 41: Ditto

utils/semver/version.go
```diff
@@ -0,0 +1,69 @@
...
+		return
+	}
+	if strings.TrimSpace(versionFilePath) == "" {
```
joshjennings98, line 44:
```suggestion
	versionFilePath := strings.TrimSpace(versionFilePath)
	if versionFilePath == "" {
```

It might better to strip it here since then it isn't stripped when you do FS.Exists later https://github.com/ARM-software/golang-utils/blob/f9cd35c9f9f09e9abb1fd3a5e7db6b59673c4105/utils/filesystem/files.go#L584

joshjennings98 (approved):
Looks good!

PR #583, opened 2025-03-17
:sparkles: `[http]` Define an HTTP Request configuration

acabarbaye:
### Description

Add configuration for HTTP request. For use when defining HTTP clients

### Test Coverage


utils/http/request.go
```diff
@@ -0,0 +1,156 @@
...
+	Scheme      string `mapstructure:"scheme"`
+	AccessToken string `mapstructure:"token"`
+}
+
+func (cfg *Auth) GetAuthorizationHeader() string {
+	if cfg.Enforced {
+		return fmt.Sprintf("%v %v", cfg.Scheme, cfg.AccessToken)
+	}
+	return ""
+}
```
joshjennings98, line 82: If you call this on an enforced but non-validated header (e.g. it has the scheme set but not the token and no-one called Validate()) you could end up with a hard to debug error as it might return a header that gets parsed quite far through the stack before having issues and you could get an error with just a blank string where it isn't clear what the issue is.

I know it isn't super nice but I think the chance of errors due to missing a call to validate is too likely. I would suggest something like:
```suggestion
const (
    missingScheme      = "!!MISSING_SCHEME!!"
    missingAccessToken = "!!MISSING_ACCESS_TOKEN!!"
)

func (cfg *Auth) GetAuthorizationHeader() string {
	if !cfg.Enforced {
	    return ""
	}
	
	scheme, accessToken := cfg.Scheme, cfg.AccessToken
	if scheme == "" {
	    scheme = missingScheme
	}
	if accessToken == "" {
	    accessToken = missingAccessToken
	}
	    
    return fmt.Sprintf("%v %v", cfg.Scheme, cfg.AccessToken)
}
```

Then if it is misused it is more obvious what the error is

joshjennings98 (dismissed):
I only had one comment that I think will help with debugging

PR #601, opened 2025-04-09
:gear: Add filesytem aware `FilePathBase()` to filepath Utils

aorabdel:
### Description
Same function as `filepath.Base` but is aware of the different file systems, mainly solves the issue of how `embed` works with `/` regardless of the OS.

Also, refactored `FilePathJoin` to work the same way.

### Test Coverage

### DIFF
--- a/utils/filesystem/filepath.go
+++ b/utils/filesystem/filepath.go
@@ -67,7 +67,7 @@ func FilePathJoin(fs FS, element ...string) string {
 	return joinedPath
 }
 
-// Same behaviour as filepath.Base, but can handle different filesystems.
+// FilePathBase has the same behaviour as filepath.Base, but can handle different filesystems.
 func FilePathBase(fs FS, fp string) string {
 	if fs == nil {
 		return ""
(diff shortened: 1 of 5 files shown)


utils/filesystem/filepath.go
```diff
@@ -54,7 +59,24 @@ func FilePathJoin(fs FS, element ...string) string {
 		return ""
 	}
 
-	return strings.ReplaceAll(filepath.Join(element...), string(platform.PathSeparator), string(fs.PathSeparator()))
```
acabarbaye, line 57: use the FilepathJoinFunction() instead

utils/filesystem/filepath.go
```diff
@@ -54,7 +59,24 @@ func FilePathJoin(fs FS, element ...string) string {
...
+	if fs.GetType() == Embed {
+		return path.Join(element...)
```
acabarbaye, line 63: if platform==fs.
filepath.Join
else{
if fs. == `/`{
path.Join
else strings.rea

acabarbaye (approved):

acabarbaye (approved):

PR #605, opened 2025-04-25
✨ Add a cache utility for files and directories

aorabdel:
### Description
Add a File Cache utility that supports multiple ways of fetching an entry and can cache both directories and individual files. Cache entries are automatically garbage-collected according to a configurable TTL.

The cache calls `IEntryRetriever.FetchEntry` to store an entry, which means this system supports multiple ways of providing an entry, like copying from another filesystem, downloading a file over HTTP, generating a file, etc.

Provide an implementation for copying a file or directory from another Filesystem to store an entry, defined in `FsEntryRetriever`

Move `CreateTestFileTree` testing utility function into a new test util package `filesystemtest`.

### Test Coverage


utils/go.mod
```diff
@@ -4,6 +4,8 @@ go 1.23.0
 
 toolchain go1.24.1
 
+replace github.com/ARM-software/golang-utils/utils/internal/testutils => ./internal/testutils
```
acabarbaye, line 7: this should not be needed

utils/internal/testutils/tmpfs.go
```diff
@@ -0,0 +1,75 @@
+package testutils
```
acabarbaye, line 1: the convention for test utils is to create a subfolder with `test` suffix of the module of interest such as `filesystemtest` in `filesystem` and add the utilities to a file called testing.go. The test util should also take a t*testing.T argument and the first statement should be t.Helper()

utils/internal/testutils/tmpfs.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// unifying timestamps
+	for _, path := range tree {
+		err = fs.Chtimes(path, fileAccessTime, fileModTime)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return tree, nil
```
acabarbaye, line 74: since it is a test util we could use require.NoError instead of returning an error

utils/simplecache/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+type ISimpleCache interface {
+	Add(context.Context, string, filesystem.FS, string) error
+	Remove(context.Context, string) error
+	Contains(context.Context, string) (bool, error)
+	Restore(context.Context, string) error
+	Close(context.Context) error
```
acabarbaye, line 14: use io.Closer

utils/simplecache/interfaces.go
```diff
@@ -0,0 +1,15 @@
+package simplecache
+
+import (
+	"context"
+
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+type ISimpleCache interface {
```
acabarbaye, line 9: could this be called filecache instead of simple cache?

utils/simplecache/config.go
```diff
@@ -0,0 +1,31 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	return validation.ValidateStruct(cfg,
+		validation.Field(&cfg.cachefs, validation.Required),
+		validation.Field(&cfg.cachePath, validation.Required),
+	)
+}
```
acabarbaye, line 31: create a func DefaultCacheConfig() Config with default values ttl and gc

utils/simplecache/config.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+
+	configUtils "github.com/ARM-software/golang-utils/utils/config"
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+type Config struct {
+	cachefs                 filesystem.FS `mapstructure:"cache_filesystem"`
```
acabarbaye, line 13: if this is configuration, field must be settable and so should be exposed

acabarbaye, line 13: this is not really settable from environment and so the filesystem should be stored as a field of the cache

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	fs "github.com/ARM-software/golang-utils/utils/filesystem"
+	"github.com/ARM-software/golang-utils/utils/parallelisation"
+)
+
+const (
```
acabarbaye, line 13: see comment about DefaultConfiguration function

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+
+type CacheEntry struct {
+	path         string
+	originalFS   fs.FS
+	originalpath string
+	expiration   time.Time
+}
+
+type Cache struct {
+	entries map[string]*CacheEntry
```
acabarbaye, line 26: the filesystem of the cache should be stored here

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	cfg     *Config
+	stopGC  context.CancelFunc
+	mu      sync.Mutex
```
acabarbaye, line 29: should be a RW mutex

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+		if now.After(entry.expiration) {
+			if err := c.cfg.cachefs.RemoveWithContext(ctx, entry.path); err == nil {
+				delete(c.entries, id)
+			}
+		}
+	}
+}
+
+func (c *Cache) isClosed() error {
+	if c.closed {
```
acabarbaye, line 48: call the Read mutex as closed is not protected

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+		return commonerrors.New(commonerrors.ErrForbidden, "cache is closed")
```
acabarbaye, line 49:
```suggestion
		return commonerrors.New(commonerrors.ErrConflict, "cache is closed")
```

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	for id, entry := range c.entries {
+		if now.After(entry.expiration) {
+			if err := c.cfg.cachefs.RemoveWithContext(ctx, entry.path); err == nil {
+				delete(c.entries, id)
+			}
+		}
+	}
+}
+
+func (c *Cache) isClosed() error {
```
acabarbaye, line 47: this look like the wrong signature isClosed() should return a boolean

utils/simplecache/cache.go
acabarbaye: Conflict is the error when state is not as expected

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	if c.closed {
+		return commonerrors.New(commonerrors.ErrForbidden, "cache is closed")
+	}
+
+	return nil
+}
+
+func (c *Cache) Contains(ctx context.Context, id string) (bool, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
acabarbaye, line 57:
```suggestion
	c.mu.RLock()
	defer c.mu.RUnlock()
```

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	err := c.cfg.cachefs.RemoveWithContext(ctx, entry.path)
+	if err != nil {
+		return err
+	}
+
+	delete(c.entries, id)
+	return nil
+}
+
+func (c *Cache) Close(ctx context.Context) error {
```
acabarbaye, line 146: this should not have a context and proceed

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+		entries: make(map[string]*CacheEntry),
+		cfg:     config,
+		stopGC:  stop,
+	}
+
+	if err := config.cachefs.MkDirAll(config.cachePath, 0755); err != nil {
+		return nil, err
+	}
+
+	parallelisation.SafeSchedule(gcCtx, cache.cfg.garbageCollectionPeriod, 0, cache.gc)
```
acabarbaye, line 207: this should be stopped on close

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+		cachefs:                 filesystem,
+		cachePath:               cachePath,
+		garbageCollectionPeriod: _DefaultGCPeriod,
+		ttl:                     _DefaultTTL,
+	}
+
+	return NewSimpleCache(ctx, config)
+}
+
+func NewSimpleCache(ctx context.Context, config *Config) (ISimpleCache, error) {
```
acabarbaye, line 181: in parallelisation there is a cancelfunc store which should be used here

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+	path         string
+	originalFS   fs.FS
+	originalpath string
+	expiration   time.Time
+}
+
+type Cache struct {
+	entries map[string]*CacheEntry
+	cfg     *Config
+	stopGC  context.CancelFunc
```
acabarbaye, line 28: use a parallelisation cancel store for htis

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+const (
+	_DefaultGCPeriod = 10 * time.Minute
+	_DefaultTTL      = 2 * time.Hour
+)
+
+type CacheEntry struct {
+	path         string
+	originalFS   fs.FS
+	originalpath string
+	expiration   time.Time
```
acabarbaye, line 22: the source may not be a file in a filesystem but something downloaded from the internet
so there should be a func() to retrieve entry

utils/simplecache/cache.go
```diff
@@ -0,0 +1,210 @@
...
+		return err
+	}
+
+	entry, exists := c.entries[id]
+	if !exists {
+		return commonerrors.Newf(commonerrors.ErrNotFound, "cache entry for %s not found", id)
+	}
+
+	srcPath := fs.FilePathJoin(c.cfg.cachefs, c.cfg.cachePath, fs.FilePathBase(c.cfg.cachefs, entry.path))
+	err := fs.CopyBetweenFS(ctx, c.cfg.cachefs, srcPath, entry.originalFS, entry.originalpath)
```
acabarbaye, line 81: this is not atomic and so there could be system trying to use the file while being copied

acabarbaye: one way to solve this is to store the file in a temp file in the same filesystem and then mv it to the new name when done. the cache should only report it has an entry if the entry is usable not in transient state

utils/simplecache/cache.go
aorabdel: it is an internal function so that the error doesn't get copied across all functions, that way if we want to change the error for any reason, we change it in one place

utils/simplecache/cache.go
aorabdel: it is, Close calls `c.stopGC()`

aorabdel: will change to CancelFunctionStore

utils/simplecache/cache.go
acabarbaye: that is fine. However, I would still change the name of the function because its current name may lead to misinterpretation is... usually returns booleans

utils/filesystem/filesystemtest/filesystemtest.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+	"github.com/ARM-software/golang-utils/utils/idgen"
+)
+
+func CreateTestFileTree(t *testing.T, fs filesystem.FS, testDir string, fileModTime time.Time, fileAccessTime time.Time) []string {
```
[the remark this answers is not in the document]
acabarbaye: since it is becoming a test helper, could you add a doc string

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+	io.Closer
+
+	// Calls the entry provider to store an entry into the cache under the provided `key`, and starts the expiration
+	// timer based on the configured TTL.
+	// Returns an error if the path doesn't exist, the key already exists in the cache, the copy process fails or
+	// the cache is closed.
+	Store(ctx context.Context, key string) error
+
+	// Evicts the cache entry identified by `key`, removing the stored cached entry immediately.
+	// Returns an error if the key does not exist in the cache, the deleting process fails or the cache is closed.
```
acabarbaye, line 23: if the key does not exist in the cache, no need to return an error. eviction is in a way successful

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+// Cache entries are identified by a key and live for a configurable TTL (time-to-live).
+// An internal garbage collector periodically evicts expired entries; the period is also configurable.
+type IFileCache interface {
+	io.Closer
+
+	// Calls the entry provider to store an entry into the cache under the provided `key`, and starts the expiration
+	// timer based on the configured TTL.
+	// Returns an error if the path doesn't exist, the key already exists in the cache, the copy process fails or
+	// the cache is closed.
+	Store(ctx context.Context, key string) error
```
acabarbaye, line 20: would it be possible to also add a method in which we can specify a ttl for a particular entry:
e.g. StoreWithTTL(ctx context, key string, ttl time.Duration)

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+	// Returns an error if the key does not exist in the cache, the copying process fails or the cache is closed.
+	Restore(ctx context.Context, key string, restoreFilesystem filesystem.FS, restorePath string) error
+}
+
+// IEntryProvider defines how to store a file or a directory into the cache.
+// Implementations might copy from a local filesystem, download over HTTP, generate a file, etc.
+type IEntryProvider interface {
+	// StoreEntry stores the entry for the given `key`` and writes it to `cacheDir` using `cacheFilesystem`.
+	// Returns the absolute path to the newly stored entry, or an error if the operation fails.
+	StoreEntry(ctx context.Context, key string, cacheFilesystem filesystem.FS, cacheDir string) (string, error)
```
acabarbaye, line 41: could this be called FetchEntry/RetrieveEntry rather than store. It does store the entry in the cache but the main part of the operation is to get the entry from wherever it comes from

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,42 @@
...
+	Evict(ctx context.Context, key string) error
+
+	// Checks whether a cache entry with the provided `key` currently exists in the cache
+	// Returns true if the entry exists, or false if it doesn't and an error if the cache is closed.
+	Has(ctx context.Context, key string) (bool, error)
+
+	// Restore copies the cached data for the provided `key` back to `restorePath` in `restoreFilesystem`.
+	// It also resets the entry's TTL, so that the most frequently accessed files remain cached.
+	// Returns an error if the key does not exist in the cache, the copying process fails or the cache is closed.
+	Restore(ctx context.Context, key string, restoreFilesystem filesystem.FS, restorePath string) error
```
acabarbaye, line 33: what is the purpose of this method? I see what it does but cannot find when it would be called

utils/filecache/fsfilecache.go
```diff
@@ -0,0 +1,36 @@
...
+type FsEntryProvider struct {
+	fs       filesystem.FS
+	basePath string
+}
+
+func (p *FsEntryProvider) StoreEntry(ctx context.Context, key string, cacheFilesystem filesystem.FS, cacheDir string) (string, error) {
+	srcPath := filesystem.FilePathJoin(p.fs, p.basePath, key)
+	if _, statErr := filesystem.Stat(srcPath); statErr != nil {
+		return "", statErr
+	}
```
acabarbaye, line 18: use filesystem.Exist as I believe this is what you are trying to find out

utils/filecache/fsfilecache.go
```diff
@@ -0,0 +1,36 @@
...
+
+	destPath := filesystem.FilePathJoin(p.fs, cacheDir, key)
+	cpErr := filesystem.CopyBetweenFS(ctx, p.fs, srcPath, cacheFilesystem, destPath)
+	if cpErr != nil {
+		return "", cpErr
+	}
+
+	return destPath, nil
+}
+
```
acabarbaye, line 28: could you add some doc string to explain what this cache is and when to use it

utils/filecache/fsfilecache.go
```diff
@@ -0,0 +1,36 @@
...
+func (p *FsEntryProvider) StoreEntry(ctx context.Context, key string, cacheFilesystem filesystem.FS, cacheDir string) (string, error) {
+	srcPath := filesystem.FilePathJoin(p.fs, p.basePath, key)
+	if _, statErr := filesystem.Stat(srcPath); statErr != nil {
+		return "", statErr
+	}
+
+	destPath := filesystem.FilePathJoin(p.fs, cacheDir, key)
+	cpErr := filesystem.CopyBetweenFS(ctx, p.fs, srcPath, cacheFilesystem, destPath)
```
acabarbaye, line 21: this is not atomic so if a different process/goroutine look for this entry in filesystem. As soon as the file is created it will consider it to be present although the copy may not be complete.

there are two ways to do so, 
- change the copyBetweenFS to be atomic but there may be some weird  corner cases to consider
- make it less prone to race conditions here
tmpKey:=fmt.sprintf("%v-tmp")
destPath := filesystem.FilePathJoin(p.fs, cacheDir, key)
destPathTmp:=filesystem.FilePathJoin(p.fs, cacheDir, tmpKey)
	cpErr := filesystem.CopyBetweenFS(ctx, p.fs, srcPath, cacheFilesystem, destPathTmp)
	if cpErr {return}
	cpErr=cacheFilesystem.Move(ctx, destPathTmp, destPath ) // This would be atomic as in the same filesystem
	if cpErr{ return}

utils/filecache/cache.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	entry, exists := c.entries[key]
+	if !exists {
+		return commonerrors.Newf(commonerrors.ErrNotFound, "cache entry for %s not found", key)
+	}
```
acabarbaye, line 138: I would not raise an error here but rather return. The eviction is indeed successful as the entry is no longer in cache

utils/filecache/cache.go
```diff
@@ -0,0 +1,206 @@
...
+
+	c.gcCancelStore.Cancel()
+
+	for id, entry := range c.entries {
+		if err := c.fs.RemoveWithContext(context.Background(), entry.path); err != nil {
+			c.closed = true
+			return commonerrors.WrapErrorf(commonerrors.ErrUnexpected, err, "could not delete %s", id)
+		}
+
+		delete(c.entries, id)
```
acabarbaye, line 165: I think record the error but try to remove as much as possible to avoid leaving the system in a weird state.
Moreover, I would potentially add something like filesystem.RmWithPrivileges(cache.cfg.CachePath) if an error happened

utils/filecache/cache.go
```diff
@@ -0,0 +1,206 @@
...
+	for id, entry := range c.entries {
+		if now.After(entry.expiration) {
+			if err := c.fs.RemoveWithContext(ctx, entry.path); err == nil {
+				delete(c.entries, id)
+			}
+		}
+	}
+}
+
+func (c *Cache) isClosed() error {
```
acabarbaye, line 42: could it be called checkIfClosed

utils/filecache/cache.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+
+	delete(c.entries, key)
+	return nil
+}
+
+func (c *Cache) Close() error {
+	if err := c.isClosed(); err != nil {
+		return err
```
acabarbaye, line 151: if close then no need to send an error. We are in the state we want to be

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// IEntryProvider defines how to store a file or a directory into the cache.
+// Implementations might copy from a local filesystem, download over HTTP, generate a file, etc.
+type IEntryProvider interface {
+	// FetchEntry fetches the entry for the given `key` and writes it to `cacheDir` using `cacheFilesystem`.
+	// Returns the absolute path to the newly stored entry, or an error if the operation fails.
+	FetchEntry(ctx context.Context, key string, cacheFilesystem filesystem.FS, cacheDir string) (string, error)
+}
```
joshjennings98, line 48: I think it would be better to do _something similar_ to this to make it easier to integrate existing fetchers/downloaders.

If we split it out then it makes it easier to extend existing objects to support the cache mechanism as we won't need to wrap everything with functions that provide a `cacheFilesystem`.

```suggestion
type IEntryProvider interface {
	// SetCache sets the filesystem `cacheFilesystem` for the entry provider
	SetCache(cacheFilesystem filesystem.FS) error
	// FetchEntry fetches the entry for the given `key` and writes it to `cacheDir`
	// Returns the absolute path to the newly stored entry, or an error if the operation fails.
	FetchEntry(ctx context.Context, key string, cacheDir string) (string, error)
}
```

utils/filecache/cache.go
```diff
@@ -0,0 +1,227 @@
...
+
+	now := time.Now()
+	for id, entry := range c.entries {
+		if now.After(entry.expiration) {
+			if err := c.fs.RemoveWithContext(ctx, entry.path); err == nil {
+				delete(c.entries, id)
+			}
+		}
+	}
+}
```
joshjennings98, line 42: There is https://pkg.go.dev/maps#DeleteFunc that can get rid of all this nesting
```suggestion
	maps.DeleteFunc(c.entries, func(id string, entry *CacheEntry) bool {
		return now.After(entry.expiration) && c.fs.RemoveWithContext(ctx, entry.path) == nil
	})
}
```

utils/filecache/cache.go
```diff
@@ -0,0 +1,227 @@
...
+	return nil
+}
+
+func (c *Cache) Has(ctx context.Context, key string) (bool, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if err := c.checkIfClosed(); err != nil {
+		return false, err
+	}
```
joshjennings98, line 61: you have locked the map already so when this calls c.mu.RLock it will block

utils/filecache/cache.go
```diff
@@ -0,0 +1,227 @@
...
+
+	_, exists := c.entries[key]
+	return exists, nil
```
joshjennings98, line 64: try to only lock for the shortest time possible, only where it is absolutely needed (i.e. map access itself) and don't lock around another function that locks.

```suggestion
func (c *Cache) Has(ctx context.Context, key string) (bool, error) {
	if err := c.checkIfClosed(); err != nil {
		return false, err
	}

	c.mu.RLock()
	defer c.mu.RUnlock()
	
	_, exists := c.entries[key]
	return exists, nil
```

utils/filecache/cache.go
```diff
@@ -0,0 +1,227 @@
...
+	defer c.mu.Unlock()
+
+	if _, exists := c.entries[key]; exists {
+		return commonerrors.Newf(commonerrors.ErrExists, "cache entry %s already exists", key)
+	}
+
+	entryPath, err := c.entryProvider.FetchEntry(ctx, key, c.fs, c.cfg.CachePath)
+	if err != nil {
+		return err
+	}
```
joshjennings98, line 90: since this doesn't touch the map I would put it before the lock to minimise how much time is spent locked

utils/filecache/cache.go
```diff
@@ -0,0 +1,227 @@
...
+
+	err = restoreFilesystem.Move(dstPath, restorePath)
+	if err != nil {
+		return err
+	}
+
+	// Sliding window approach, which means frequently used items will stay cached
+	entry.expiration = time.Now().Add(entry.ttl)
+
```
joshjennings98, line 134: this locks for a long time doing possibly long-running filesystem actions.

I think it needs refactoring to minimise the time locked. I think it is fine to lock multiple times in the same function if it minimises the lock time total. We definitely shouldn't lock whilst doing copies and things because that would block the entire cache

utils/filecache/cache.go
joshjennings98: my mistake, they are read locks so you can have multiple

utils/filecache/entrymap.go
```diff
@@ -0,0 +1,42 @@
+package filecache
+
+import "sync"
+
+type entryMap struct {
+	entries sync.Map
```
joshjennings98, line 6: can you justify the use of sync.Map vs a map with a mutex.

Since sync.Map is not recommended to be used unless there are specific conditions, I think it needs a comment explaining why it is used.

See https://pkg.go.dev/sync#Map

utils/filecache/entrymap.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (em *entryMap) Load(key string) ICacheEntry {
+	entry, _ := em.entries.Load(key)
+	return entry.(ICacheEntry)
```
joshjennings98, line 11: entry can be nil so this might cause a panic

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
+package filecache
```
acabarbaye, line 1: could file cache be a submodule of a top module being either filesystem or cache

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// IFileCache provides a filesystem cache for files and directories.
+// Cache entries are identified by a key and live for a configurable TTL (time-to-live).
+// An internal garbage collector periodically evicts expired entries; the period is also configurable.
+type IFileCache interface {
+	io.Closer
+
+	// Calls the entry provider to store an entry into the cache under the provided `key`, and starts the expiration
+	// timer based on the configured TTL.
```
acabarbaye, line 18:
```suggestion
	// timer based on the configured cache overall TTL.
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// Returns an error if the key already exists in the cache, the action of fetching an entry fails or the cache is closed.
+	Store(ctx context.Context, key string) error
+
+	// StoreWithTTL calls the entry provider to store an entry into the cache under the provided `key`, and starts the expiration
```
acabarbaye, line 22:
```suggestion
	// StoreWithTTL stores an entry into the cache under the provided `key`, and starts the expiration
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// timer using the specified `ttl` (overriding the default if non-zero).
+	// Returns an error if the key already exists in the cache, the action of fetching an entry fails or the cache is closed.
+	StoreWithTTL(ctx context.Context, key string, ttl time.Duration) error
+
+	// Evicts the cache entry identified by `key`, removing the stored cached entry immediately.
```
acabarbaye, line 27:
```suggestion
	// Evict removes the cache entry identified by `key`.
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// Returns an error if the deleting process fails or the cache is closed.
+	Evict(ctx context.Context, key string) error
+
+	// Checks whether a cache entry with the provided `key` currently exists in the cache
```
acabarbaye, line 31:
```suggestion
	// Has checks whether a cache entry with the provided `key` currently exists in the cache
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// Returns true if the entry exists, or false if it doesn't.
+	// Returns an error if the action of looking up the entry resulted in an error or the cache is closed.
+	Has(ctx context.Context, key string) (bool, error)
+
+	// Fetch copies the cached data for the provided `key` back to `destPath` in `destFilesystem`.
```
acabarbaye, line 36:
```suggestion
	// Fetch copies the cached data for the provided `key` to the destination `destPath` in `destFilesystem`.
```

utils/filecache/entrymap.go
joshjennings98, line 1: If you really want to use the sync.Map then we need to be careful since it doesn't have the same type safety as a normal map (see all the `any` types)

Because it is in the same package as all of the rest of the file cache code, the fields that are private to the entrymap are accessible to things using it (e.g. `Cache`).

This means you can do: `cache.entries.entries.Store(any, any)` and bypass the methods that you added for safety. 

If you really want to use sync.Map then I think it should be in a sub package of this package. Then only the public values will be exposed and there isn't a way to use the entries unsafely

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// It also resets the entry's TTL, so that the most frequently accessed files remain cached.
+	// Returns an error if the key does not exist in the cache, the copying process fails or the cache is closed.
+	Fetch(ctx context.Context, key string, destFilesystem filesystem.FS, destPath string) error
+}
+
+// IEntryProvider defines how to store a file or a directory into the cache.
```
acabarbaye, line 42:
```suggestion
// IEntryRetriever defines how to retrieve a file or a directory from a particular source in order to store it temporarily into a cache.
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+// Implementations might copy from a local filesystem, download over HTTP, generate a file, etc.
+type IEntryProvider interface {
```
acabarbaye, line 44:
```suggestion
type IEntryRetriever interface {
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	Has(ctx context.Context, key string) (bool, error)
+
+	// Fetch copies the cached data for the provided `key` back to `destPath` in `destFilesystem`.
+	// It also resets the entry's TTL, so that the most frequently accessed files remain cached.
+	// Returns an error if the key does not exist in the cache, the copying process fails or the cache is closed.
+	Fetch(ctx context.Context, key string, destFilesystem filesystem.FS, destPath string) error
+}
+
+// IEntryProvider defines how to store a file or a directory into the cache.
+// Implementations might copy from a local filesystem, download over HTTP, generate a file, etc.
```
acabarbaye, line 43:
```suggestion
// Implementations might copy from a local filesystem, download over HTTP, generate a file, etc and the method of retrieval is defined by the FetchEntry method.
```

utils/filecache/cache.go
```diff
@@ -0,0 +1,200 @@
+package filecache
+
+import (
+	"context"
+	"errors"
+	"sync/atomic"
+	"time"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	fs "github.com/ARM-software/golang-utils/utils/filesystem"
```
joshjennings98, line 10: It is best to avoid renaming imports unless you have to. It isn't a rule but it is recommended by a lot of people.

There are a few cases where it is recommended but I think replacing a descriptive name with 2 letters is not one of them.

This google style guide is helpful https://google.github.io/styleguide/go/decisions.html#import-renaming

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+type IEntryProvider interface {
+	// SetCacheFilesystem sets the cache filesystem for the entry provider
+	SetCacheFilesystem(fs filesystem.FS) error
+	// SetCacheDir sets the base cache directory for the entry provider
+	SetCacheDir(dir string) error
+	// FetchEntry fetches the entry for the given `key` and writes it to `cacheDir` using `cacheFilesystem`.
```
acabarbaye, line 49:
```suggestion
	// FetchEntry fetches the entry for the given `key` and stores it into the cache.
```

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	// Returns the absolute path to the newly stored entry, or an error if the operation fails.
+	FetchEntry(ctx context.Context, key string) (string, error)
+}
+
+type ICacheEntry interface {
```
acabarbaye, line 54: could you add some doc strings

utils/filecache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+	entries       *entryMap
+	entriesLM     *lockMap
+	entryProvider IEntryProvider
+	fs            fs.FS
+	cfg           *FileCacheConfig
+	cancelStore   *parallelisation.CancelFunctionStore
+	closed        atomic.Bool
+}
+
+func (c *Cache) gc(ctx context.Context, _ time.Time) {
```
joshjennings98, line 24: it is probabaly a good idea to check the context when you loop through this since it could be a long running operation. use parallelisation.DetermineContextError (or whatever it is called) so you can break early if the context is cancelled

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	Copy(ctx context.Context, cacheFs filesystem.FS, destFs filesystem.FS, destPath string) error
```
acabarbaye, line 55: I believe the cacheFS should not need specifying here

utils/filecache/cache.go
joshjennings98: also there should be a test that this works

utils/filecache/interfaces.go
```diff
@@ -0,0 +1,59 @@
...
+	Delete(ctx context.Context, fs filesystem.FS) error
```
acabarbaye, line 56: why does an fs needs specifying

utils/filecache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+	c.entries.Range(func(key string, entry ICacheEntry) {
+		if entry.IsExpired() && c.entriesLM.TryLock(key) {
+			if err := entry.Delete(ctx, c.fs); err == nil {
+				c.entries.Delete(key)
+				c.entriesLM.Unlock(key)
```
joshjennings98, line 29: It is good to use defer for unlocks etc. It might be fine now, but in the event we add more logic it could cause issues

```suggestion
			defer c.entriesLM.Unlock(key)
			if err := entry.Delete(ctx, c.fs); err == nil {
				c.entries.Delete(key)
```

utils/filecache/cache.go
```diff
@@ -0,0 +1,200 @@
...
+func (c *Cache) Close() error {
+	if err := c.checkIfClosed(); err != nil {
+		return nil
+	}
+
+	c.closed.Store(true)
+	c.entriesLM.Clear()
+	c.cancelStore.Cancel()
+
+	ctx := context.Background()
```
joshjennings98, line 134: this close function may not be cancellable. since you use context.Background here, we cannot every cancel it.

I know it may not be something you don't want to cancel since it is cleaning up, but when we run things in kubernetes etc. they use signals to shutdown pods etc.

If a pod was being deleted then this cleanup would block and cause the pod to be stuck terminating which might cause issues, especially since if we were removing a pod or something we would be getting rid of the filesystem data anyway

utils/filecache/fsfilecache.go
```diff
@@ -0,0 +1,78 @@
...
+		return commonerrors.New(commonerrors.ErrInvalid, "the cache filesystem cannot be nil")
+	}
+
+	p.cachefs = fs
+
+	return nil
+}
+
+func (p *FsEntryProvider) SetCacheDir(dir string) error {
+	if exists := p.cachefs.Exists(dir); !exists {
```
joshjennings98, line 54: just simplify this, it isn't a complex condition

```suggestion
	if !p.cachefs.Exists(dir) {
```

utils/filecache/fsfilecache.go
```diff
@@ -0,0 +1,78 @@
...
+	cachefs       filesystem.FS
+	basePath      string
+	cacheBasePath string
+}
+
+// FetchEntry locates the resource identified by `key` within the provider’s source filesystem under `basePath`,
+// copies it into `cacheDir` using `cacheFilesystem`, and returns the absolute path of the cached entry.
+func (p *FsEntryProvider) FetchEntry(ctx context.Context, key string) (string, error) {
+	srcPath := filesystem.FilePathJoin(p.fs, p.basePath, key)
+	if exists := p.fs.Exists(srcPath); !exists {
```
joshjennings98, line 23: just simplify this, it isn't a complex condition

```suggestion
	if !p.fs.Exists(srcPath) {
```

utils/cache/filecache/entrymap.go
```diff
@@ -0,0 +1,74 @@
...
+type iEntryMap interface {
+	Load(key string) ICacheEntry
+	Store(key string, entry ICacheEntry)
+	Range(f func(key string, entry ICacheEntry))
+	Exists(key string) bool
+	Delete(key string)
+	Clear()
+}
+type entryMap struct {
+	// entries holds a map of per-key cache entry.
```
joshjennings98, line 16: thanks for adding this :+1:

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,210 @@
...
+	if err := parallelisation.DetermineContextError(ctx); err != nil {
+		return
+	}
+
+	c.entries.Range(func(key string, entry ICacheEntry) {
+		if entry.IsExpired() && c.entriesLM.TryLock(key) {
+			if err := entry.Delete(ctx); err == nil {
+				c.entries.Delete(key)
+				defer func() {
+					c.entriesLM.Unlock(key)
```
joshjennings98, line 34: what happens if an entry fails the delete, this won't get run. Then every tryLock will fail to lock.

joshjennings98: This is an example of where mocking would be good because you can have an explicit entry.Delete failure and see the behaviour and verify it won't happen

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,210 @@
...
+	})
+}
+
+func (c *Cache) checkIfClosed() error {
+	if c.closed.Load() {
+		return commonerrors.New(commonerrors.ErrConflict, "cache is closed")
+	}
+
+	return nil
+}
```
joshjennings98, line 48:
```suggestion
func (c *Cache) closed() bool {
	return c.closed.Load()
}
```

I don't think this should return an error type. To me, and error implied an issue occurred when the function ran, but in here the entire point of the function is to check if the cache is closed.

so if the cache is closed then the function has done it's job correctly and it shouldn't return an error.

I would then return an error when you do things like:

```go
    if c.closed() {
		return false, commonerrors.New(commonerrors.ErrConflict, "cache is closed")
	}
	```

File: utils/cache/filecache/filecache.go [hunk 63]
```diff
@@ -0,0 +1,210 @@
...
+	if err := c.checkIfClosed(); err != nil {
+		return err
+	}
+
+	if ttl == 0 {
+		return commonerrors.New(commonerrors.ErrInvalid, "TTL for an entry can't be zero")
+	}
+
+	if c.entries.Exists(key) {
+		return commonerrors.Newf(commonerrors.ErrExists, "cache entry %s already exists", key)
```
joshjennings98, line 72:
```suggestion
		return commonerrors.Newf(commonerrors.ErrExists, "cache entry '%s' already exists", key)
```

joshjennings98: This applied everywhere in this file.

It is a good practice to wrap them in quotes as if they are unset then you can tell by the error because it will say something like `entry '' already set` so you can tell the issue is that it is empty.

You can even do `%q` which will automatically add the quotes. Adrien likes to use `%v` for all of them just to keep it simple though.

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,210 @@
...
+
+	if len(removalErrors) > 0 {
+		err := c.fs.RemoveWithPrivileges(ctx, c.cfg.CachePath)
+		if err != nil {
+			removalErrors = append(removalErrors, commonerrors.WrapErrorf(commonerrors.ErrUnexpected, err, "could not delete %s", c.cfg.CachePath))
+			return errors.Join(removalErrors...)
+		}
+
+		c.entries.Clear()
+	}
```
joshjennings98, line 167: what is the point of this?

you call RemoveWithPrivileges on the whole cache of there was any error. Why not just do this to start with instead of the c.entries.Range?

utils/cache/filecache/filecache_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+			var cacheTree []string
+			err = cacheFs.ListDirTree(tmpCacheDir, &cacheTree)
+			require.NoError(t, err)
+
+			var srcTree []string
+			err = srcFs.ListDirTree(tmpSrcDir, &srcTree)
+			require.NoError(t, err)
+
+			require.Equal(t, len(srcTree), len(cacheTree), "Cache Dir and Src Dir have differnet number of content")
```
joshjennings98, line 83:
```suggestion
			require.Equal(t, len(srcTree), len(cacheTree), "Cache Dir and Src Dir have different number of content")
```

utils/cache/filecache/filecache_test.go
```diff
@@ -0,0 +1,754 @@
...
+		require.NoError(t, err)
+		require.False(t, filesystem.Exists(tmpfilePath), "could not remove the src file")
+
+		err = cache.Fetch(ctx, tmpfileBase, fs, tmpfilePath)
+		require.NoError(t, err)
+
+		require.True(t, filesystem.Exists(tmpfilePath), "cache did not fetch the file")
+	})
+
+	t.Run("Fetch_file_non_existent", func(t *testing.T) {
```
joshjennings98, line 190: Just so you're aware, you should make the test names human readable in the source code. 

When we run the tests it is go that will handle the underscores and stuff.

utils/cache/filecache/filecache.go
aorabdel: this is so that we clean as much as we can, first we try to delete one entry at a time skipping failures, then if there were any failures we try to delete the whole thing with privileges, if that failed then we display those errors, this sequence will maximize the number of thing we clean, if RemoveWithPrivileges fails for some reason, then we haven't cleaned anything

utils/cache/filecache/filecache_test.go
```diff
@@ -0,0 +1,754 @@
...
+		require.NoError(t, err)
+		require.Equal(t, originalFileContent, cacheFileContent)
+	})
+}
+
+func TestFileCache_Remove(t *testing.T) {
+	t.Run("Remove_file", func(t *testing.T) {
+		defer goleak.VerifyNone(t)
+
+		fs := filesystem.NewFs(filesystem.StandardFS)
```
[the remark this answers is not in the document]
joshjennings98: filesystem.NewFs is better if you wrap the tests in a loop that tests on the different filesystems we support like https://github.com/ARM-software/golang-utils/blob/c71c52a9f448237b03be3abfa4b8871aec29c876/utils/filesystem/files_test.go#L38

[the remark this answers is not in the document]
joshjennings98: not a criticism, just some information

utils/cache/filecache/filecache_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+	t.Run("Restore_Evict", func(t *testing.T) {
+		defer goleak.VerifyNone(t)
+
+		fs := filesystem.NewFs(filesystem.StandardFS)
+		tmpCacheDir, err := fs.TempDirInTempDir("test-cache")
+		require.NoError(t, err)
+		defer func() { _ = fs.Rm(tmpCacheDir) }()
+
+		tmpSrcDir, err := fs.TempDirInTempDir("test-cache-src")
```
joshjennings98, line 684: in tests it is better to use `t.TmpDir()` to create temporary directories.

this lets the test suite itself handle the cleanup and you won't be left with any files if your defer function fails.

I think you should make this change everywhere

utils/cache/filecache/filecache_test.go
```diff
@@ -0,0 +1,754 @@
...
+
+		err = g.Wait()
+		require.NoError(t, err)
+
+		actualHash, err := hasher.CalculateFile(fs, tmpTestFilePath)
+		require.NoError(t, err)
+
+		require.Equal(t, actualHash, expectedHash, "the copied file got corrupted")
+	})
+}
```
joshjennings98, line 754: Could you add a test that does more than two go routines. Like something that checks it works with 50 go routines?

If it works with 50 then we know the concurrency will probably be fine.

Two go routines is barely concurrent and if they aren't actually running at the same time then we aren't even testing the concurrency. With 50 it is less likely that they will not be running at the same time.

joshjennings98: I only say this to be thorough. The point of this code is to do something concurrent and therefore we really need to test a bunch of concurrency

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,210 @@
...
+	"time"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+	"github.com/ARM-software/golang-utils/utils/parallelisation"
+)
+
+type Cache struct {
+	entries       iEntryMap
+	entriesLM     iLockMap
```
joshjennings98, line 16: thanks for using the interfaces

utils/cache/filecache/interfaces.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+)
+
+// IFileCache provides a filesystem cache for files and directories.
+// Cache entries are identified by a key and live for a configurable TTL (time-to-live).
+// An internal garbage collector periodically evicts expired entries; the period is also configurable.
+type IFileCache interface {
+	// Store stores an entry into the cache under the provided `key`, and starts the expiration
+	// timer based on the configured cache overall TTL.
+	// Returns an error if the key already exists in the cache, the action of fetching an entry fails or the cache is closed.
```
joshjennings98, line 16: the doc strings are used to generate documentation public documentation https://pkg.go.dev/github.com/ARM-software/golang-utils/utils. This is automatic and not controlled by us.

It is good practice to write them in a way that makes sense when read as documentation.

It would make more sense to says "It returns XXX" rather than "Returns XXX"

utils/cache/filecache/filecache_test.go
aorabdel: There is a test called "Caches" that covers this, this is to test a single per-key lock that is guaranteed to overlap

utils/cache/filecache/filecache.go
joshjennings98: if the goal is to remove all of the cache, then why not just do that in the first place instead of removing each individual file?

joshjennings98: If remove with privileges fails then it doesn't really matter how many we cleaned up as there is a bigger issue

utils/cache/filecache/filecache.go
aorabdel: the purpose of the function is to wrap the error, if this need to change, then we have one place to maintain this, this error is supposed to be in every cache operation, that why I wanted to refactor that in a separate function

utils/cache/filecache/filecache.go
joshjennings98: either do this or just try to delete the files then leave the ones that couldn't be deleted and log it.

joshjennings98: or fall back to remove with privileges if the normal remove fails but do it per file. that would achieve the same thing.

utils/cache/filecache/filecache_test.go
joshjennings98: oh okay :+1:

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,208 @@
...
+		return
+	}
+
+	c.entries.Range(func(key string, entry ICacheEntry) {
+		if entry.IsExpired() && c.entriesLM.TryLock(key) {
+			if err := entry.Delete(ctx); err == nil {
+				c.entries.Delete(key)
+				defer c.entriesLM.Delete(key)
+			}
+			defer c.entriesLM.Unlock(key)
```
joshjennings98, line 39: move that to just after the tryLock, otherwise it won't get run if the if statement isn't entered

joshjennings98, line 39: It is a good idea to always put a defer close just after the thing that you opened. Then if any code changes it doesn't move and it is easy to keep track of and see that the resource is freed

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,208 @@
...
+	"time"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/filesystem"
+	"github.com/ARM-software/golang-utils/utils/logs/logrimp"
+	"github.com/ARM-software/golang-utils/utils/parallelisation"
+	"github.com/ARM-software/golang-utils/utils/retry"
+)
+
+var closedErr = commonerrors.New(commonerrors.ErrConflict, "cache is closed")
```
joshjennings98, line 16: This is good!

utils/cache/filecache/filecache.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil, commonerrors.WrapError(commonerrors.ErrInvalid, err, "invalid configuration")
+
+	}
+
+	if entryRetriever == nil {
+		return nil, commonerrors.New(commonerrors.ErrInvalid, "the entry provider cannot be nil")
+	}
+
+	if cacheFilesystem == nil {
+		return nil, commonerrors.New(commonerrors.ErrInvalid, "the cache filesystem cannot be nil")
```
joshjennings98, line 181: use `commonerrors.Undefined` for undefined variables

utils/cache/filecache/interfaces.go
joshjennings98, line 1: thanks for improving the docstrings

joshjennings98 (approved):
Just two very minor things, then feel free to merge it

utils/cache/filecache/interfaces.go
aorabdel: no problem

PR #616, opened 2025-05-16
:sparkles: `[collection]` Add utilities to serialising maps into comma separated strings

acabarbaye:
### Description

Added a few utilities to convert maps into slices or strings so that they can be serialised easily

### Test Coverage

### DIFF
--- a/utils/collection/parseLists.go
+++ b/utils/collection/parseLists.go
@@ -77,6 +77,10 @@ func ParseCommaSeparatedListToMap(input string) (pairs map[string]string, err er
 
 // ParseCommaSeparatedListOfPairsToMap returns a map of key value pairs from a string containing a comma separated list of pairs using pairSeparator to separate between keys and values e.g. key1=value1,key2=value2
 func ParseCommaSeparatedListOfPairsToMap(input, pairSeparator string) (pairs map[string]string, err error) {
+	if pairSeparator == "," {
+		pairs, err = ParseCommaSeparatedListToMap(input)
+		return
+	}
 	inputSplit := ParseCommaSeparatedList(input)
 	pairs = make(map[string]string, len(inputSplit))
 	for i := range inputSplit {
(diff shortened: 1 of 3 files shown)


utils/collection/parseLists.go
```diff
@@ -73,3 +74,69 @@ func ParseCommaSeparatedListToMap(input string) (pairs map[string]string, err er
...
+		default:
+			err = commonerrors.Newf(commonerrors.ErrInvalid, "could not parse key value pair '%v'", inputSplit[i])
+			return
+		}
+	}
+	return
+}
+
+// ConvertSliceToCommaSeparatedList converts a slice into a string containing a coma separated list
+func ConvertSliceToCommaSeparatedList[T comparable](slice []T) string {
```
aorabdel, line 98: I think `Stringer` would be a better choice for this use case

aorabdel: same for all comparable use cases

utils/collection/parseLists.go
```diff
@@ -73,3 +74,69 @@ func ParseCommaSeparatedListToMap(input string) (pairs map[string]string, err er
 
 	return
 }
+
+// ParseCommaSeparatedListOfPairsToMap returns a map of key value pairs from a string containing a comma separated list of pairs using pairSeparator to separate between keys and values e.g. key1=value1,key2=value2
+func ParseCommaSeparatedListOfPairsToMap(input, pairSeparator string) (pairs map[string]string, err error) {
```
aorabdel, line 79: This won't work if the `pairSeparator` is a comma, we'd need to two inputs
`listSeparator` : separates pairs - `pairSeparator` separates key and value,  then do special handling for when the two separators are equal

utils/collection/parseLists.go
acabarbaye: this is a good point and I decided to go one step further and accept anything so that built-in types are also accepted (they do not implement Stringer)

aorabdel (approved):

PR #622, opened 2025-05-30
:sparkles: Generic map and slice serialisers

acabarbaye:
### Description

- extended mapstructure decode
- add ways to flatten and expand maps

### Test Coverage


utils/maps/map.go
```diff
@@ -0,0 +1,88 @@
...
+func (m Map) AsMap() map[string]string {
+	return m
+}
+
+// Delete deletes a key out of the maps with the given prefix.
+func (m Map) Delete(prefix string) {
+	for k := range m {
+		match := k == prefix
+		if !match {
+			if !strings.HasPrefix(k, prefix) {
```
joshjennings98, line 37: HasPrefix works on a sting if it is entirely prefix so you can remove the match I think

joshjennings98 (approved):
Just 2 comments, otherwise I think it is okay

utils/maps/map.go
acabarbaye: I used the code as is from terraform and only made some small changes. I will make changes as part of further developments

joshjennings98 (approved):

PR #624, opened 2025-05-30
:sparkles: `subprocess` Add support for terminating processes instead of killing them

joshjennings98:
### Description

Add support for terminating processes instead of killing them

### Test Coverage


utils/subprocess/executor.go
```diff
@@ -268,6 +268,13 @@ func (s *Subprocess) Stop() (err error) {
 	return s.stop(true)
 }
 
+// Interrupt stops the process via the TERM signal.
```
acabarbaye, line 271:
```suggestion
// Interrupt terminates the process
```

acabarbaye (approved):

utils/subprocess/executor.go
acabarbaye: remember TERM Signal means nothing on windows

PR #634, opened 2025-06-06
✨ `subprocess` Add interrupt with context and interrupt signal option 

joshjennings98:
### Description

Add interrupt with context and interrupt signal option

### Test Coverage

### DIFF
--- a/utils/proc/ps_posix.go
+++ b/utils/proc/ps_posix.go
@@ -18,13 +18,23 @@ import (
 	"github.com/ARM-software/golang-utils/utils/parallelisation"
 )
 
+func getpgid(pid int) (gpid int, err error) {
+	gpid, err = syscall.Getpgid(pid)
+	if err != nil {
+		err = commonerrors.WrapErrorf(commonerrors.ErrUnexpected, err, "could not get pgid of '%v'", pid)
+		return
+	}
+
+	return
+}
+
 func killGroup(ctx context.Context, pid int32) (err error) {
 	err = parallelisation.DetermineContextError(ctx)
 	if err != nil {
 		return
 	}
 	// see https://varunksaini.com/posts/kiling-processes-in-go/
-	pgid, err := syscall.Getpgid(int(pid))
+	pgid, err := getpgid(int(pid))
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


utils/subprocess/executor.go
```diff
@@ -192,12 +196,24 @@ func (s *Subprocess) IsOn() bool {
...
-// Wait waits for the command to exit and waits for any copying to
-// stdin or copying from stdout or stderr to complete.
-//
-// The command must have been started by Start.
-func (s *Subprocess) Wait() error {
-	return s.command.cmdWrapper.cmd.Wait()
+// Wait waits for the command to stop existing on the system.
+// This allows check to work if the underlying process was stopped.
+func (s *Subprocess) Wait(ctx context.Context) (err error) {
+	pid := s.command.cmdWrapper.cmd.Process.Pid
```
acabarbaye, line 202: process could be nil so do a similar check to https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/os/exec/exec.go;drc=7e0950861703c93ea3e7f122c1c594851919261e;l=916

utils/parallelisation/wait.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+)
+
+type IErrorWaiter interface {
+	Wait() error
+}
+
```
acabarbaye, line 12: could there be some doc strings to understand when and why to use it

utils/proc/ps_posix.go
```diff
@@ -36,3 +37,18 @@ func killGroup(ctx context.Context, pid int32) (err error) {
 	err = ConvertProcessError(syscall.Kill(-pgid, syscall.SIGKILL))
 	return
 }
+
+// WaitForCompletion will wait for a given process to complete.
+// This allows check to work if the underlying process was stopped without needing the os.Process that started it.
+func WaitForCompletion(ctx context.Context, pid int) (err error) {
+	pgid, err := syscall.Getpgid(pid)
```
acabarbaye, line 44: is there not a generic proc FindPGId or such a function in the process library in use uner the bonnet. If so, could we use it instead

acabarbaye: and could there be a helper for it?

acabarbaye (approved):
I have a couple of comments but overall looks great. will need to run it on Windows to verify it works fully there but don't wait for me to integrate

utils/subprocess/executor.go
```diff
@@ -192,12 +193,17 @@ func (s *Subprocess) IsOn() bool {
...
+// This allows check to work if the underlying process was stopped.
+func (s *Subprocess) Wait(ctx context.Context) (err error) {
+	var pid int
+	if s.command != nil && s.command.cmdWrapper.cmd != nil && s.command.cmdWrapper.cmd.Process != nil {
+		pid = s.command.cmdWrapper.cmd.Process.Pid
+	} else {
+		return commonerrors.New(commonerrors.ErrCondition, "command not started")
+	}
+
+	return proc.WaitForCompletion(ctx, pid)
```
acabarbaye, line 206: just so that we know behaviour is not broken on windows. Could you have s temporary //FIXME if platform.isWindows{ call the go wait method as it used to be }

PR #669, opened 2025-08-07
:sparkles: `commonerrors` Add ErrFailed

joshjennings98:
### Description

Add ErrFailed. This is useful because ErrUnexpected is used a lot as a system error even when we are carrying out underlying tasks where a failure isn't an error. This provides us another generic error but one that we can use to distinguish between failures and system errors. 

### Test Coverage

### DIFF
--- a/utils/commonerrors/errors.go
+++ b/utils/commonerrors/errors.go
@@ -45,7 +45,11 @@ var (
 	ErrEOF                = errors.New("end of file")
 	ErrMalicious          = errors.New("suspected malicious intent")
 	ErrOutOfRange         = errors.New("out of range")
-	ErrFailed             = errors.New("failed")
+	// ErrFailed should be used as a generic error where an error is an expected and valid state.
+	// For example a failing command may cause subproccess.Execute to return an error if the command exits with 1 but
+	// this wouldn't be a system error and you might want to distinguish between this and the subprocess wrapper erroring
+	// when you pass the message up the stack.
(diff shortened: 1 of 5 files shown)


utils/commonerrors/errors.go
```diff
@@ -45,6 +45,7 @@ var (
 	ErrEOF                = errors.New("end of file")
 	ErrMalicious          = errors.New("suspected malicious intent")
 	ErrOutOfRange         = errors.New("out of range")
+	ErrFailed             = errors.New("failed")
```
acabarbaye, line 48: could you add some docstring to mention when to use this.

aorabdel:
Do we want to use this in proc/error `ConvertProcessError` ?

utils/commonerrors/errors.go
```diff
@@ -312,7 +319,8 @@ func WrapError(targetError, originalError error, msg string) error {
 		if cleansedMsg == "" {
 			return New(tErr, originalError.Error())
 		} else {
-			return Errorf(tErr, "%v%v %v", cleansedMsg, string(TypeReasonErrorSeparator), originalError.Error())
+			return Errorf(
+				tErr, "%v%v %v", cleansedMsg, string(TypeReasonErrorSeparator), originalError)
```
acabarbaye, line 323:
```suggestion
				tErr, "%v%v %v", cleansedMsg, string(TypeReasonErrorSeparator), originalError.Error())
```

PR #703, opened 2025-09-09
:sparkles: `safeio` Add support for cancelling readers that make blocking kernel reads during copying

joshjennings98:
### Description

Add support for cancelling readers that make blocking kernel reads during copying to safeio submodule.

When copying, a lot of the time things were in the kernel read and therefore since the cancel in the other contextual readers only cancel before a read, they weren't be cancelled.

This allows readclosers to be used and makes it so that context cancellation will call their close functions which will stop the kernel read.

https://man7.org/linux/man-pages/man2/read.2.html

### Test Coverage

### DIFF
--- a/utils/safeio/read.go
+++ b/utils/safeio/read.go
@@ -105,7 +105,7 @@ func NewContextualReadCloser(ctx context.Context, reader io.ReadCloser) io.ReadC
 		reader: contextio.NewReader(ctx, reader),
 		close: func() error {
 			_ = stop()
-			return reader.Close()
+			return nil
 		},
 		closed: atomic.NewBool(false),
 	}
--- a/utils/safeio/read.go
+++ b/utils/safeio/read.go
@@ -6,6 +6,7 @@ import (
 	"io"
 
 	"github.com/dolmen-go/contextio"
+	"go.uber.org/atomic"
 
 	"github.com/ARM-software/golang-utils/utils/commonerrors"
 	"github.com/ARM-software/golang-utils/utils/parallelisation"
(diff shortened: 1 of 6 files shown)


utils/safeio/read.go
```diff
@@ -76,6 +77,42 @@ func NewContextualReader(ctx context.Context, reader io.Reader) io.Reader {
...
+// Context state is checked during the read and close is called if the context is cancelled
+// This allows for readers that block on syscalls to be stopped via a context
+func NewContextualReadCloser(ctx context.Context, reader io.ReadCloser) io.ReadCloser {
+	stop := context.AfterFunc(ctx, func() { _ = reader.Close() })
+
+	r := safeReadCloser{
+		reader: contextio.NewReader(ctx, reader),
+		close: func() error {
+			_ = stop()
+			return reader.Close()
```
acabarbaye, line 108:
```suggestion
       _ = stop()
		return  nil
```

utils/safeio/read.go
```diff
@@ -76,6 +77,42 @@ func NewContextualReader(ctx context.Context, reader io.Reader) io.Reader {
 	return contextio.NewReader(ctx, reader)
 }
 
+type safeReadCloser struct {
+	reader io.Reader
```
acabarbaye, line 81: mention this makes it idempotent

utils/safeio/read_closer_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewContextualReadCloser(t *testing.T) {
+	t.Run("Normal contextual reader blocks even after cancel", func(t *testing.T) {
+		r, w, err := os.Pipe()
```
acabarbaye, line 16: goroutine leak check should be added

PR #872, opened 2026-05-21
:sparkles: Added modules and helpers for carrying out operations on YAML and JSON

acabarbaye:
### Description

- added validation rules for paths
- added standard ways for serialising JSON and YAML content/files
- added a standard way to validate a JSON or YAML file against a JSON Schema

### Test Coverage


utils/filesystem/rules.go
```diff
@@ -0,0 +1,174 @@
...
+
+// NewOSPathExistRule returns a validation rule to use in configuration.
+// The rule checks whether a string is a valid path for the operating system's
+// filesystem and actually exists.
+// `when` describes whether the rule is enforced or not.
+func NewOSPathExistRule(when bool) validation.Rule {
+	return NewPathExistRule(GetGlobalFileSystem(), when)
+}
+
+type pathExistValidationRule struct {
```
joshjennings98, line 99: why can't the other type be reused?

why not pull out the validate() and pass it as part of the constructor rather than duplicating the type every time we have a new rule?

It could be like how `http.HandlerFunc` and `http.ServeHTTP` works

utils/serialization/json/helpers.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/pquerna/ffjson/ffjson"
+	sigsyaml "sigs.k8s.io/yaml"
+
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+)
+
+var nullBytes = []byte("null")
+
+func isNilInterface(i any) bool {
+	return (*[2]uintptr)(unsafe.Pointer(&i))[1] == 0
```
joshjennings98: is this really the best way to do this? Why isn't it handled by our existing reflection functions

utils/serialization/json/helpers.go
```diff
@@ -0,0 +1,104 @@
...
+package json
+
+import (
+	"bytes"
+	"context"
+	"unsafe"
+
+	"github.com/mailru/easyjson"
+	"github.com/pquerna/ffjson/ffjson"
+	sigsyaml "sigs.k8s.io/yaml"
```
joshjennings98, line 21: It might be worth directly using https://github.com/yaml/go-yaml instead of the wrapper. As far as I know we only use sigs.k8s.io/yaml to convert between JSON and YAML but now we have pulled out this stuff into here, maybe it makes sense to add the helper here.

Then we aren't reliant on them keeping sigs.k8s.io/yaml up to date and https://github.com/yaml/go-yaml is managed by the official YAML organisation. And as far as I can tell the sigs.k8s.io/yaml just uses it under the hood anyway

utils/serialization/yaml/helpers.go
joshjennings98, line 1: This would be where we get rid of sigsyaml I guess

utils/validation/jsonschema/validator.go
joshjennings98, line 1: is it worth adding ILimits to these? Or at least an option for with ILimits?

it might get messy if you try to do this with a very large file or request

joshjennings98 (dismissed):
I think this looks good. I only had a few comments

utils/filesystem/rules.go
acabarbaye: I defined the type once but this change was just about moving already defined code for ease of finding it. I agree there should be helpers for defining validation rules although ozzo-valdiation  define a lot of them

utils/serialization/json/helpers.go
acabarbaye: now done. This is so it is a faster check

utils/serialization/json/helpers.go
acabarbaye: This is a fair point but will be done as part of a separate PR. the goal was to provide helpers to do this and use what we are currently using in our codebase but I agree I should be providing our own helper implementation to reduce the dependencies and reuse our parsing helpers

utils/validation/jsonschema/validator.go
acabarbaye: good point and I have done this now on the schema and on the file to validate against

joshjennings98 (approved):

PR #886, opened 2026-05-29
:bug: Ensure that there is an explicit check that a path is not a directory before unlinking

joshjennings98:
### Description

Ensure that there is an explicit check that a path is not a directory before unlinking.

The check for an error with `"is a directory"` (`EISDIR`) is not robust enough. On MacOS, the error is `ERPERM` which means that `Remove` fails tests on MacOS.

### Test Coverage

### DIFF
--- a/utils/filesystem/extendedosfs.go
+++ b/utils/filesystem/extendedosfs.go
@@ -22,14 +22,14 @@ type ExtendedOsFs struct {
 }
 
 func (c *ExtendedOsFs) Remove(name string) (err error) {
-	info, err := c.OsFs.Stat(name)
+	info, err := c.Stat(name)
 	if err != nil {
 		err = ConvertFileSystemError(err)
 		return
 	}
 
 	if !info.IsDir() {
-		// The following is to ensure sockets are correctly removed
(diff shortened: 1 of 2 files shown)


utils/filesystem/extendedosfs.go
```diff
@@ -22,14 +22,21 @@ type ExtendedOsFs struct {
 }
 
 func (c *ExtendedOsFs) Remove(name string) (err error) {
-	// The following is to ensure sockets are correctly removed
-	// https://stackoverflow.com/questions/16681944/how-to-reliably-unlink-a-unix-domain-socket-in-go-programming-language
-	err = commonerrors.Ignore(ConvertFileSystemError(syscall.Unlink(name)), commonerrors.ErrNotFound)
-	err = commonerrors.IgnoreCorrespondTo(err, "is a directory")
+	info, err := c.Stat(name)
 	if err != nil {
+		err = ConvertFileSystemError(err)
```
acabarbaye, line 27: if error is notFound the error should be ignored

acabarbaye (changes requested):

PR #888, opened 2026-06-01
:bug: Ensure that if a file starts with a byte-order mark that it is still parsable

joshjennings98:
### Description

Ensure that if a file starts with a byte-order mark that it is still parsable.

Some tools add invisible byte-order marks to files that are invisible to users but will break parsers that expect the frontmatter to be the first part of the file

https://en.wikipedia.org/wiki/Byte_order_mark

:warning: This doesn't fix the problem, it just provides some tests that show it :warning:

### Test Coverage

- [ ]  This change is covered by existing or additional automated tests.


changes/20260601172808.bugfix
```diff
@@ -0,0 +1 @@
+:bug: Ensure that if a file starts with a byte-order mark that it is still parsable
```
acabarbaye:
```suggestion
:bug: `[frontmatter]` Ensure that if a file starts with a byte-order mark that it is still parsable
```

acabarbaye (changes requested):

acabarbaye (approved):

PR #892, opened 2026-06-02
:bug: Ensure that different formats with the same start fence are treated as distinct formats

joshjennings98:
### Description

Ensure that different formats with the same start fence are treated as distinct formats

the old parser would:
1. see ---
2. pick the first --- format
3. look only for closing ---
4. hit EOF / fail
5. report not found if it had a different end

This will still fallback to one the first parser (like it did before) if you have formats that aren't compatible e.g.:

Start: "---", End: "---", include delimiters false
Start: "---", End: "---", include delimiters true

### Test Coverage

### DIFF
--- a/changes/20260602103158.bugfix
+++ b/changes/20260602103158.bugfix
@@ -1,0 +1,1 @@
+:bug: Ensure that different formats with the same start fence are treated as distinct formats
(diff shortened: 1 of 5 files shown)


changes/20260602103158.bugfix
```diff
@@ -0,0 +1 @@
+:bug: Ensure that different formats with the same start fence are treated as distinct formats
```
acabarbaye:
```suggestion
:bug: `[frontmatter]` Ensure that different formats with the same start fence are treated as distinct formats
```

acabarbaye (approved):

PR #898, opened 2026-06-08
:bug: `[collection]` Fixed collection sequence helpers so nil `iter.Seq` values no longer panic

acabarbaye:
### Description
range on a nil sequence panics. This is to avoid this

### Test Coverage


utils/collection/stack/stack.go
```diff
@@ -73,6 +75,9 @@ func (s *Stack[T]) Push(value ...T) {
 }
 
 func (s *Stack[T]) PushSequence(seq iter.Seq[T]) {
+	if seq == nil {
```
joshjennings98: Why not use sequencyOrEmpty here?

joshjennings98 (approved):

PR #904, opened 2026-06-09
:sparkles: `[collection]` Add helpers such as `First...` `Last...`

acabarbaye:
### Description

further helpers on collection 

### Test Coverage

### DIFF
--- a/utils/collection/access.go
+++ b/utils/collection/access.go
@@ -1,11 +1,23 @@
 package collection
 
-import "iter"
+import (
+	"iter"
+	"slices"
+)
 
 // At returns the element at index from slice.
 //
+// Negative indices are interpreted relative to the end of the slice, following
+// the same pattern as Python indexing, so `-1` selects the last element, `-2`
+// the penultimate one, and so on.
+//
 // If index is out of range, it returns the zero value and false.
+//
+// References:
+//   - https://docs.python.org/3/tutorial/introduction.html#lists
+//   - https://www.geeksforgeeks.org/python/what-is-negative-indexing-in-python/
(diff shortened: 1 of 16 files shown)


utils/collection/first_last.go
```diff
@@ -0,0 +1,108 @@
...
+
+// FirstSequence returns the first element yielded by seq.
+func FirstSequence[E any](seq iter.Seq[E]) (element E, ok bool) {
+	for v := range SequenceOrEmpty(seq) {
+		element = v
+		ok = true
+		return
+	}
+	return
+}
```
joshjennings98, line 31:
```suggestion
func FirstSequence[E any](seq iter.Seq[E]) (element E, ok bool) {
	return AtSequence(seq, 0)
}
```

utils/collection/access.go
```diff
@@ -0,0 +1,43 @@
+package collection
+
+import "iter"
+
+// At returns the element at index from slice.
+//
+// If index is out of range, it returns the zero value and false.
+func At[S ~[]E, E any](slice S, index int) (element E, ok bool) {
+	if index < 0 || index >= len(slice) {
```
joshjennings98, line 21: Not saying it needs it, but it would be cool if negative indexes worked in reverse like python

joshjennings98: Then we could use this in the `Last` like my suggestion for `First`

joshjennings98 (dismissed):
Just some minor suggestions

utils/collection/access.go
acabarbaye: this was a great suggestion and so looked at python builtin helpers as further inspiration. I added a few more

joshjennings98 (approved):

PR #908, opened 2026-06-18
:sparkles: Added a package to change string casing

acabarbaye:
This is so that https://go.dev/wiki/CodeReviewComments#initialisms are correctly handled

### DIFF
--- a/utils/strings/casing/replacement.go
+++ b/utils/strings/casing/replacement.go
@@ -115,7 +115,7 @@ func (r *Replacer) WriteString(ctx context.Context, w io.Writer, s string) (n in
 }
 
 func splitCamelWords(value string) []string {
-	if value == "" {
+	if reflection.IsEmpty(value) {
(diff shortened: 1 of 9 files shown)


utils/strings/casing/replacement.go
```diff
@@ -0,0 +1,174 @@
...
+	return builder.String()
+}
+
+// WriteString rewrites s with Replace and writes the result to w using the repository's context-aware safe I/O helpers.
+func (r *Replacer) WriteString(ctx context.Context, w io.Writer, s string) (n int, err error) {
+	return safeio.WriteString(ctx, w, r.Replace(s))
+}
+
+func splitCamelWords(value string) []string {
+	if value == "" {
```
Kem-Gov:
```suggestion
	if reflection.IsEmpty(value) {
```

Kem-Gov (approved):

PR #923, opened 2026-07-02
:sparkles: `[validation]` Added JSON-Schema-oriented validation helpers for numbers, strings, arrays, objects

acabarbaye:
This is creating the most common rules we use in our different projects. It uses [ozzo-validation](https://github.com/go-ozzo/ozzo-validation) but provides some helpers which should match most JSON schema validation like rules we use and so should reduce duplication


utils/validation/collection_rules.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/ARM-software/golang-utils/utils/collection"
+)
+
+// ArrayItems validates that every item in an array or slice satisfies rule.
+//
+// Example: `ArrayItems(Type("string"))` accepts `[]any{"a", "b"}`.
+func ArrayItems(rule validation.Rule) validation.Rule {
+	return validation.By(func(value any) error {
+		items, err := typedSequence[any](value)
+		if err != nil || items == nil {
```
joshjennings98, line 19: This function verifies against slices, but if you pass in something that is nil then `typedSequence` returns `nil, nil` (due to the call to `validation.Indirect`) and so the validation rule passes. I would expect it to reject non arrays/slices

e.g.:

```go
var s *string
assert.Error(t, validation.Validate(s, ArrayItems(Type("string"))))

var f func() = nil
assert.Error(t, validation.Validate(f, ArrayItems(Type("string"))))

var m map[int]string
assert.Error(t, validation.Validate(m, ArrayItems(Type("string"))))
```

utils/validation/collection_rules.go
```diff
@@ -0,0 +1,71 @@
...
+// Example: `MapKeys(Pattern(regexp.MustCompile("^[a-z]+$")))` accepts
+// `map[string]any{"alpha": 1}`.
+func MapKeys(rule validation.Rule) validation.Rule {
+	return PropertyNames(rule)
+}
+
+// MapValues validates that every value in a map satisfies rule.
+//
+// Example: `MapValues(Type("string"))` accepts `map[string]any{"a": "x"}`.
+func MapValues(rule validation.Rule) validation.Rule {
```
joshjennings98: same for both this and `mapKeys`, it does not reject non-maps

utils/validation/collection_rules_test.go
```diff
@@ -0,0 +1,44 @@
...
+func TestCollectionRules(t *testing.T) {
+	t.Run("array items", func(t *testing.T) {
+		assert.NoError(t, validation.Validate([]any{"a", "b"}, ArrayItems(Type("string"))))
+		assert.Error(t, validation.Validate([]any{"a", 1}, ArrayItems(Type("string"))))
+
+		seq := iter.Seq[any](func(yield func(any) bool) {
+			_ = yield("a")
+			_ = yield("b")
+		})
+		assert.NoError(t, validation.Validate(seq, ArrayItems(Type("string"))))
```
joshjennings98, line 32: add explicit tests here for the above comments (even if you decide it is expected behavior)

utils/validation/collection_rules.go
acabarbaye: good spot and covering those cases

utils/validation/collection_rules_test.go
acabarbaye: done

utils/validation/collection_rules.go
acabarbaye: I have done the same

acabarbaye: and added explicit tests

utils/validation/time_rules.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// IsDuration validates whether a string or byte slice is a valid Go duration.
+var IsDuration = validation.NewStringRule(func(value string) bool {
+	_, err := time.ParseDuration(value)
+	return err == nil
+}, "must be a valid duration")
+
+// IsTimestamp validates whether a string or byte slice is a valid RFC3339 timestamp.
+var IsTimestamp = validation.NewStringRule(func(value string) bool {
```
joshjennings98: it might be worth includibg RFC3339 in the rule name just because there are other timestamps in `time` and it isn't obvious without reading the docstring

acabarbaye: fair. changing it

utils/validation/utils.go
```diff
@@ -0,0 +1,498 @@
...
+			if recovered := recover(); recovered != nil {
+				switch err := recovered.(type) {
+				case error:
+					callErr = err
+				default:
+					panic(recovered)
+				}
+			}
+		}()
+		rv.Call([]reflect.Value{yield})
```
joshjennings98: if you pass anything that panics during the call for a separate runtime issue, this will be caught in the recover and treated as a validation error, even if it should remain a panic.

utils/validation/string_rules_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"regexp"
+	"testing"
+
+	validation "github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStringRules(t *testing.T) {
```
joshjennings98, line 14: can you add some expected errors when trying to validate things that aren't strings? Currently even the assert.Errors are happy path tests and I think that since we are creating custom rules we should be sure that they behave as expected (and validation.Validate take an interface{} so it could take anything)

utils/validation/json-schema-rules_test.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/go-ozzo/ozzo-validation/v4/is"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/ARM-software/golang-utils/utils/collection"
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/commonerrors/errortest"
+	"github.com/ARM-software/golang-utils/utils/field"
+)
+
+func TestJSONSchemaInspiredRules(t *testing.T) {
```
joshjennings98, line 23: same comment about non-happy path tests

utils/validation/logical.go
```diff
@@ -248,3 +254,278 @@ func NewNoneRuleWithContext(rule ...validation.RuleWithContext) validation.RuleW
...
+// It is a readability-oriented alias for [Exactly].
+func NOf(want int, rule ...validation.Rule) validation.Rule {
+	return Exactly(want, rule...)
+}
+
+// NOfWithContext returns a context-aware rule that succeeds if exactly want of
+// the provided context-aware rules succeed.
+//
+// It is a readability-oriented alias for [ExactlyWithContext].
+func NOfWithContext(want int, rule ...validation.RuleWithContext) validation.RuleWithContext {
```
joshjennings98: what is with the naming of these ones, what is NOf, to me it seems like an inverse of Exactly?

acabarbaye: this is N of something

utils/validation/logical.go
```diff
@@ -248,3 +258,318 @@ func NewNoneRuleWithContext(rule ...validation.RuleWithContext) validation.RuleW
...
+func ExactlyWithContext(n int, rule ...validation.RuleWithContext) validation.RuleWithContext {
+	c := NewExactlyCompositeRule(n)
+	c.AppendContextualRule(rule...)
+	return c
+}
+
+// NOf returns a rule that succeeds only if exactly `n` of the provided rules
+// succeed.
+//
+// In other words, this is an “N of these rules must pass” combinator.
```
joshjennings98, line 390: thanks for adding this, `N of` makes a lot more sense than how I interpreted as `No f`

utils/validation/time_rules.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// IsDuration validates whether a string or byte slice is a valid Go duration.
+var IsDuration = validation.NewStringRule(func(value string) bool {
+	_, err := time.ParseDuration(value)
+	return err == nil
+}, "must be a valid duration")
+
+// IsRFC3339Timestamp validates whether a string or byte slice is a valid RFC3339 timestamp.
+var IsRFC3339Timestamp = validation.NewStringRule(func(value string) bool {
```
joshjennings98, line 192: thanks for this

utils/reflection/reflection.go
```diff
@@ -118,6 +119,105 @@ func SetStructField(structure any, fieldName string, value any) error {
...
+	stringKey := reflect.ValueOf(key)
+	if stringKey.Type().AssignableTo(keyType) {
+		return stringKey, true
+	}
+	if stringKey.Type().ConvertibleTo(keyType) {
+		return stringKey.Convert(keyType), true
+	}
+	if keyType.Kind() == reflect.Interface {
+		return stringKey, true
+	}
```
joshjennings98: This can panic when you have interface keyed maps:

```
type key string

func (k key) String() string {
	return string(k)
}

value := map[fmt.Stringer]any{
	key("a"): 1,
}
```

If you did this contrived example then the following will panic (it should fail because `"a"` doesn't satisfy `fmt.Stringer`:

```
validation.Validate(value, RequiredProperties("a"))
```

BEcause of this line:

```
if keyType.Kind() == reflect.Interface {
	return stringKey, true
}
```

so it turns it into a string which then panics when you call MapIndex because a normal string doesn't satisfy `fmt.Stringer`.

So we need to remove the following:
```suggestion
```

because `stringKey.Type().AssignableTo(keyType)` correctly rejects the string as not implementing `fmt.Stringer` but then the `keyType.Kind() == reflect.Interface` overrides that because `fmt.Stringer` is an interface. 

Removing it won't be an issue because the earlier cehcks already cover every safe conversion:

- `map[string]...`: `AssignableTo` succeeds.
- `map[MyString]`...: `ConvertibleTo` succeeds.
- `map[any]`...: `AssignableTo` succeeds.
- `map[fmt.Stringer]`...: both fail, correctly, because plain string is incompatible.

joshjennings98 (approved):
Thanks for addressing the comments, this looks good

PR #927, opened 2026-07-08
:bug: `[casing]` fix casing helpers so adjacent acronym runs like AESRSA are preserved in existing camelCase and PascalCase identifiers instead of collapsing to Aesrsa.

acabarbaye:
This is a bug in the casing helpers as this edge case was not considered


joshjennings98 (approved):
Thanks for finding and fixing this issue

PR #938, opened 2026-07-16
:sparkles: `[url]` Add helpers for normalising and comparing URLs

acabarbaye:
Comparing URLs is actually quite difficult as they have many encoding possible, many elements which may or may not be important: See https://www.rfc-editor.org/info/rfc3986/#section-6 for more details

### DIFF
--- a/utils/url/canonical.go
+++ b/utils/url/canonical.go
@@ -1,0 +1,661 @@
+package url
+
+import (
+	"net"
+	netUrl "net/url"
+	stdpath "path"
+	"slices"
+	"strings"
+
+	"github.com/ARM-software/golang-utils/utils/collection"
+	"github.com/ARM-software/golang-utils/utils/commonerrors"
+	"github.com/ARM-software/golang-utils/utils/reflection"
+	"golang.org/x/net/idna"
+)
+
+var defaultSchemePorts = map[string]string{
+	"ftp":   "21",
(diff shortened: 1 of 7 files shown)


utils/url/canonical.go
```diff
@@ -0,0 +1,662 @@
...
+//	match, err := CompareURLs("https://www.example.com/path", "https://example.com/path", RemoveWWW())
+//	// match == true
+//
+//	match, err = CompareURLs("https://api.example.com/v1/users", "/v1/users", IgnoreScheme(), IgnoreHost())
+//	// match == true
+//
+// References:
+//   - RFC 3986: https://datatracker.ietf.org/doc/html/rfc3986
+//   - URL normalisation overview: https://en.wikipedia.org/wiki/URL_normalization
+func CompareURLs(url1, url2 string, options ...NormalisationOption) (match bool, err error) {
```
joshjennings98: It might be useful to have a version of this that has the signature needed for `SortFunc` so it can be used in slice utilities

joshjennings98 (dismissed):

utils/url/canonical.go
acabarbaye: I created such function

joshjennings98 (dismissed):

joshjennings98 (approved):

PR #943, opened 2026-07-16
:sparkles: Added some helpers for generating summaries

acabarbaye:
Extend loggers so we can generate summaries


utils/logs/summary/base.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+	"strings"
+
+	baselogs "github.com/ARM-software/golang-utils/utils/logs"
+)
+
+var _ ISummaryLogger = &baseSummaryLogger{}
+
+// baseSummaryLogger adapts a standard logger pair to the summary logger
```
joshjennings98: if it is for adapting a logger to a summary logger, would it make sense to add a constructor instead of building it like `baseSummaryLogger{l}` etc.

utils/logs/summary/file.go
```diff
@@ -0,0 +1,29 @@
...
+func NewFileSummaryLogger(path string, loggerSource string) (logger *FileSummaryLogger, err error) {
+	if reflection.IsEmpty(path) {
+		err = commonerrors.UndefinedVariable("log file path")
+		return
+	}
+	bLogger, err := baselogs.NewTextFileOnlyLogger(path, loggerSource)
+	if err != nil {
+		return
+	}
+	logger = &FileSummaryLogger{baseSummaryLogger{
```
joshjennings98: to avoid this, it just seems a bit nasty. I know it is unlikely that we will modify the struct but if we did then we would have to update it everywhere we do this vs just updating the logger

joshjennings98 (dismissed):
Just a comment about using a constructor

utils/logs/summary/file.go
acabarbaye: not sure I fully understood what you meant

joshjennings98 (approved):
thanks for the changes

